PR #7, opened 2014-09-05
Make additional claims configurable

niftyn8:


samdengler:
Implemented variation on this in #7

PR #10, opened 2015-04-18
fix no expiration option error

deraru:
Hi

I have detected `NoMethodError (undefined method expiration for #<Doorkeeper::OpenidConnect::Config:0x007ff85333c4c8>):` error.
So I have added expiration option with default value 120 seconds.

If you have any suitable defaults, could you please inform me it?

Thanks


deraru:
:confetti_ball:

Issue #12, opened 2015-05-07
IdToken#as_jws_token sets an invalid kid


gilbertquri:
According to the JWT spec:

The "kid" (key ID) header parameter is a hint indicating which specific key owned by the signer should be used to validate the signature. This allows signers to explicitly signal a change of key to recipients. Omitting this parameter is equivalent to setting it to an empty string. The interpretation of the contents of the "kid" parameter is unspecified. This header parameter is OPTIONAL.

I don't think setting it to the public key is the right thing to do. Specifically, to be interoperable with mod_auth_openidc it would be better to leave out the kid parameter.

samdengler:
@gilbertquri thanks for the clarification.  It might be a couple of days before I can dig into this.  If you have an approach in mind, could you submit a PR?  Thanks!

gchang10:
I think you can just leave out the kid parameter when encoding the id token. I will submit a PR shortly.

samdengler:
Thanks!

gilbertquri:
I don't seem to have permission to push up a branch for creating a PR. Can you please add me?

samdengler:
Hey @gilbertquri - we're using the Fork & Pull strategy (https://help.github.com/articles/using-pull-requests/#fork--pull)  Can you fork the repo and submit the PR from there?

samdengler:
Here's a link to the spec for future reference if needed: http://self-issued.info/docs/draft-jones-json-web-token-01.html#ReservedHeaderParameterName

PR #14, opened 2015-05-08
Leave out the kid parameter from the id token headers

gchang10:
Usually a token recipient will either have the public cert locally or call the JWKS endpoint to retrieve the public cert if the OP supports it.


samdengler:
closes #12

PR #15, opened 2015-05-15
Check the presence of id_token when rendering it in the response body

gilbertquri:
This will prevent breaking grant types that are not openid connect enabled such as client_credentials.

### DIFF
--- a/lib/doorkeeper/openid_connect.rb
+++ b/lib/doorkeeper/openid_connect.rb
@@ -57,7 +57,7 @@ module Doorkeeper
 
       def body
         original_body.
-          merge({:id_token => id_token.as_jws_token}).
+          merge({:id_token => id_token.try(:as_jws_token)}).
           reject { |_, value| value.blank? }
       end
     end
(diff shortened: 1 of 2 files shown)


gilbertquri:
@samdengler Any comment on this PR?

samdengler:
@gilbertquri This looks good, I'll merge it.

I don't think I'm actually enabling openid connect correctly according to the spec (http://openid.net/specs/openid-connect-core-1_0.html?rfc.section.3.1.2.1#AuthRequest), which says: "OpenID Connect requests MUST contain the openid scope value."

If that were working, I don't think we need the fix you've supplied.

I think I have to coordinate with the Doorkeeper project perhaps to make request data available to this project.  I'll try and find some time on how to do that.

gilbertquri:
The purpose of the check is actually for compatibility with pure OAuth grant flows. For example, the client_credentials flow (which does not generate an id token) will continue to work with the presence of your gem.

nbibler:
Along the same lines, a token refresh flow appears to error similarly to the client_credential flow without this Pull. 👍

PR #26, opened 2016-12-21
Support configurable protocol.

gigr:
Heya! Was going to make an issue for this to discuss the implementation, but the change was so small I figured I might as well associate it with a PR.

In a nutshell, this makes the `protocol` used in the discovery endpoints configurable. We use `https` even in `development` and without this configuration option we have to monkey patch `Doorkeeper::OpenidConnect::DiscoveryController#protocol`.

### DIFF
--- a/lib/doorkeeper/openid_connect/config.rb
+++ b/lib/doorkeeper/openid_connect/config.rb
@@ -125,7 +125,6 @@ module Doorkeeper
 
       def protocol
         @protocol ||= ::Rails.env.production? ? :https : :http
-        @protocol
       end
     end
   end
--- a/lib/doorkeeper/openid_connect/config.rb
+++ b/lib/doorkeeper/openid_connect/config.rb
@@ -121,7 +121,11 @@ module Doorkeeper
 
       option :claims, builder_class: ClaimsBuilder
 
-      option :protocol, default: ::Rails.env.production? ? :https : :http
+      option :protocol
+
+      def protocol
+        @protocol ||= ::Rails.env.production? ? :https : :http
+      end
     end
   end
 end
--- a/lib/doorkeeper/openid_connect/config.rb
+++ b/lib/doorkeeper/openid_connect/config.rb
@@ -117,6 +117,8 @@ module Doorkeeper
       option :expiration, default: 120
 
       option :claims, builder_class: ClaimsBuilder
+
+      option :protocol, default: ::Rails.env.production? ? :https : :http
     end
   end
 end
--- a/lib/doorkeeper/openid_connect/config.rb
+++ b/lib/doorkeeper/openid_connect/config.rb
@@ -125,6 +125,7 @@ module Doorkeeper
 
       def protocol
         @protocol ||= ::Rails.env.production? ? :https : :http
+        @protocol
       end
     end
   end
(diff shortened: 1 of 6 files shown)


toupeira:
Hey @gigr, thanks this looks great! Just two small requests:

- please add 1-2 examples to `spec/lib/config_spec.rb` to test the default values for `protocol`
- rebase and squash everything into one commit using our [commit message conventions](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/CONTRIBUTING.md#commit-message-conventions), e.g. something like "feat: Support configurable protocols" would be great

gigr:
Totally blanked on config specs! Will do!

gigr:
Looking even further, is there a reason that we explicitly set the protocol? Why not just use whatever the Rails default is configured to?

[This commit](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/commit/2618e043) suggests we did it to conform to the spec, but isn't it better to ensure we're publishing URIs that the application can handle?

lib/doorkeeper/openid_connect/config.rb
```diff
@@ -120,6 +120,12 @@ def extended(base)
       option :expiration, default: 120
 
       option :claims, builder_class: ClaimsBuilder
+
+      option :protocol
+
+      def protocol
+        @protocol ||= ::Rails.env.production? ? :https : :http
+      end
```
gigr: This is the only way that I could figure to provide a default. Using the `default` option means that the value is only set when `lib/openid_connect/config` is required. In practice this should be fine, but it's also hard to test.

gigr:
Will squash once the final review is done.

toupeira:
> Looking even further, is there a reason that we explicitly set the protocol? Why not just use whatever the Rails default is configured to?
>
> This commit suggests we did it to conform to the spec, but isn't it better to ensure we're publishing URIs that the application can handle?

It was indeed to conform to the spec, OIC mandates HTTPS and OAuth2 in general should also always be encrypted.

The problem is that, as far as I know, Rails doesn't know if HTTPS is available or not, there's `default_url_options` but that might still be set to `protocol: :http` if an app only has some protected routes.

I think it makes sense to default to HTTPS and require a deliberate configuration change to use HTTP, I'll probably tweak the README a bit after this is merged to make it more obvious.

lib/doorkeeper/openid_connect/config.rb
```diff
@@ -120,6 +120,13 @@ def extended(base)
       option :expiration, default: 120
 
       option :claims, builder_class: ClaimsBuilder
+
+      option :protocol
+
+      def protocol
+        @protocol ||= ::Rails.env.production? ? :https : :http
+        @protocol
```
toupeira: Second line is unnecessary, `||=` already returns the left-hand side.

gigr:
@toupeira Should be good to go now. Thanks for your help!

gigr:
@toupeira I fixed merged conflicts. Good to go?

toupeira:
@gigr there are still pending review comments :) If you want I can take over and will credit you in the CHANGELOG.

gigr:
Oh shoot! Sorry totally missed those. Will fix up today.

toupeira:
@gigr I want to make a new release so I went ahead and merged my complaints into #32, the commit is still yours though and you're also credited in the CHANGELOG :) Thanks again!

PR #27, opened 2016-12-27
Send scopes to claim generators

nbibler:
It is occasionally useful to know _which_ scopes a calling application is using when generating claim values.

In the current implementation, `instance_eval` is used which causes unexpected context changes (`self` is the resource owner instance within the claim block which is undocumented and unexpected). Ruby's `instance_eval` also disallows additional arguments to be passed to the block.

Rather than using `instance_eval` this just `call`s the block with the resource owner and new scopes argument. Thereby sending the caller's scopes and leaving the block context unchanged.

💥  Undocumented breaking change: The context within the `normal_claim` is no longer the resource owner instance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * The configuration setting `jws_public_key` wasn't actually used, it's deprecated now and will be removed in the next major release
 * nil values and empty strings are now removed from the UserInfo and IdToken responses
-
+* Claims now receive an optional second `scopes` argument which allow you to dynamically adjust claim values based on the requesting applications' scopes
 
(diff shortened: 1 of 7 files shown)


nbibler:
Also: If you are open to this change, I'd like to replace another couple of `instance_eval` calls in the library in other PRs. For example, it's surprising that I can manipulate token instances from within `resource_owner_from_access_token` and `subject`'s scope is the resource owner.

nbibler:
CHANGELOG and README entries added.

toupeira:
Thanks! I'll release a 1.1.1 soon after merging one or two of the other open PRs.

PR #28, opened 2016-12-28
Fix standard claim mapping for claims with symbolic names

tylerhunt:
Both the examples and the initializer template show claims being defined with symbols for names, but the standard claims are defined as strings, so they'll never map this way. 

``` ruby
claim('email', &:email) # sets scope to :email (correct)
claim(:email, &:email) # sets scope to :profile (incorrect)
```

I've made the most unobtrusive change possible to fix this, but let me know if it'd be better to `to_s`/`to_sym` the name and/or scope on assignment instead.


toupeira:
@tylerhunt thanks for the PR and catching this bug!

I think we should convert both `@name` and `@scope` to symbols in `Claim#initialize` and use `%i` in `STANDARD_CLAIMS` so they're symbols instead of strings.

Let me know if you have time to look into this, otherwise I'll change it myself and will credit you in the CHANGELOG.

tylerhunt:
@toupeira Pushed up the changes you requested.

PR #29, opened 2017-01-06
Add mongoid5 orm support

alechoey:


toupeira:
Thanks for the PR, was there a specific reason for closing it?

This gem currently only uses ActiveRecord to temporarily store nonces, I've already wondered if it would make more sense to let the user define how to save/load nonces in the initializer, or maybe even use something like the `Rails.cache` interface.

PR #30, opened 2017-01-14
Loosen json-jwt dependency restriction to ~> 1.6.

nbibler:
This allows developers to use the current [v1.7.0 json-jwt](https://github.com/nov/json-jwt/tree/v1.7.0) release which appears to pass all current tests.


toupeira:
Thanks!

PR #31, opened 2017-01-15
Use call rather than instance_eval for configuration blocks

nbibler:
This removes some of the surprising context changes within the configuration blocks. Specifically, `self` is no longer the caller instance. This was an undocumented and surprising side effect of using `instance_eval` to execute blocks.


toupeira:
Thanks again :)

PR #33, opened 2017-04-17
feature: Support pairwise subject identifier

110y:
According to [OpenID Connect specification](http://openid.net/specs/openid-connect-core-1_0.html#SubjectIDTypes), we can use `pairwise` subject identifier that provides a different subject identifier to each client.
This pull request adds abillity for generating pairwise subject identifier to this gem.

How about this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,21 @@ The following settings are required in `config/initializers/doorkeeper_openid_co
 - `subject`
   - Identifier for the resource owner (i.e. the authenticated user). A locally unique and never reassigned identifier within the issuer for the end-user, which is intended to be consumed by the client. The value is a case-sensitive string and must not exceed 255 ASCII characters in length.
   - The database ID of the user is an acceptable choice if you don't mind leaking that information.
+  - If you want to provide a different subject identifier to each client, use [pairwise subject identifier](http://openid.net/specs/openid-connect-core-1_0.html#SubjectIDTypes) with configurations like below.
+
+    ```ruby
+    # config/initializers/doorkeeper_openid_connect.rb 
+    Doorkeeper::OpenidConnect.configure do
+    # ...
+      subject_types_supported [:pairwise]
+
+      subject do |resource_owner, application|
+        Digest::SHA256.hexdigest("#{resource_owner.id}#{URI.parse(application.redirect_uri).host}#{'your_secret_salt'}")
+      end
+    # ...
+    end
+    ```
+
 - `jws_private_key`
   - Private RSA key for [JSON Web Signature](https://tools.ietf.org/html/draft-ietf-jose-json-web-signature-31).
   - You can generate a private key with the `openssl` command, see e.g. [Generate a keypair using OpenSSL](https://en.wikibooks.org/wiki/Cryptography/Generate_a_keypair_using_OpenSSL).
--- a/lib/doorkeeper/openid_connect/config.rb
+++ b/lib/doorkeeper/openid_connect/config.rb
@@ -93,6 +93,7 @@ module Doorkeeper
 
       option :jws_private_key
       option :issuer
+      option :subject_types_supported, default: [:public]
 
       option :resource_owner_from_access_token, default: lambda { |*_|
         fail Errors::InvalidConfiguration, I18n.translate('doorkeeper.openid_connect.errors.messages.resource_owner_from_access_token_not_configured')
--- a/lib/generators/doorkeeper/openid_connect/templates/initializer.rb
+++ b/lib/generators/doorkeeper/openid_connect/templates/initializer.rb
@@ -7,6 +7,8 @@ Doorkeeper::OpenidConnect.configure do
 -----END RSA PRIVATE KEY-----
 EOL
 
+  subject_types_supported [:public]
+
   resource_owner_from_access_token do |access_token|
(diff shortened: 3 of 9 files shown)


110y:
@toupeira 
Do you have any thought on this PR?

toupeira:
Hey Yuko, thanks for this PR! I'm currently on an extended vacation so
unfortunately I don't have time for a review until around middle of June.


On 17 May 2017 18:26, "Yuki Ito" <notifications@github.com> wrote:

@toupeira <https://github.com/toupeira>
Do you have any thought on this PR?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/33#issuecomment-302061864>,
or mute the thread

lib/generators/doorkeeper/openid_connect/templates/initializer.rb
```diff
@@ -24,9 +26,12 @@
...
   end
 
-  subject do |resource_owner|
+  subject do |resource_owner, application|
     # Example implementation:
-    # resource_owner.key
+    # resource_owner.id
+
+    # or if you need pairwise subject identifier, implement like below:
+    # Digest::SHA256.hexdigest("#{resource_owner.id}#{URI.parse(application.redirect_uri).host}#{'your_secret_salt'}")
```
toupeira: I don't think it's a good idea to reuse the same lambda for the pairwise algorithm, and 'm wondering if we need to support both pairwise and non-pairwise algorithms at the same time.

From the [specs](http://openid.net/specs/openid-connect-core-1_0.html#SubjectIDTypes):

> The OpenID Provider's Discovery document SHOULD list its supported Subject Identifier types in the `subject_types_supported` element. If there is more than one type listed in the array, the Client MAY elect to provide its preferred identifier type using the `subject_type` parameter during Registration. 

If I understand this correctly, the `subject_type` is entirely determined by the server side and can't be requested by the client, and we currently don't support dynamic registration or have a `subject_type` attribute on `oauth_applications` so we don't necessarily need to support both simultaneously, but it might be better in preperation to adding registration in the future.

What do you think about adding a second setting called `pairwise_subject`, and then make `subject_types_supported` dynamic, i.e. check if `subject` and `pairwise_subject` are defined and return the corresponding keys?

toupeira:
@110y thanks for this PR and sorry again for the delay, I had a look through and added some comments!

lib/generators/doorkeeper/openid_connect/templates/initializer.rb
110y: > the Client MAY elect to provide its preferred identifier type using the subject_type parameter during Registration.

According to this description, we must implement features for [Open ID Connect Dynamic Client Registration](https://openid.net/specs/openid-connect-registration-1_0.html) to add ability for Client election of subject types, right?

If so, I think we should create another PR to cover multiple subject types since its implementation isn't trivial.

What's your opinion on this?  @toupeira

README.md
```diff
@@ -87,6 +87,7 @@ The following settings are required in `config/initializers/doorkeeper_openid_co
 - `subject`
   - Identifier for the resource owner (i.e. the authenticated user). A locally unique and never reassigned identifier within the issuer for the end-user, which is intended to be consumed by the client. The value is a case-sensitive string and must not exceed 255 ASCII characters in length.
   - The database ID of the user is an acceptable choice if you don't mind leaking that information.
+  - If you want to provide a different subject identifier to each client, use [pairwise subject identifier](http://openid.net/specs/openid-connect-core-1_0.html#SubjectIDTypes).
```
110y: fixed d773579

lib/doorkeeper/openid_connect/config.rb
```diff
@@ -93,6 +93,7 @@ def extended(base)
 
       option :jws_private_key
       option :issuer
+      option :subject_types_supported
```
110y: fixed d773579

110y:
@toupeira fixed some issues. Please check again 🙇

lib/generators/doorkeeper/openid_connect/templates/initializer.rb
toupeira: Hmm ok I see now that it's not as straightforward as I thought, since in `IdToken#subject` we have no way of knowing which algorithm we should use (if we had both a `subject` and a `pairwise_subject`).

My concern is breaking configuration in future releases, and also users setting `subject_types_supported [:pairwise]` but not actually implementing a pairwise subject algorithm, but both of these are not real dealbreakers. Let me know if you see another solution, otherwise I think we can merge this.

One last request: Can you squash the commits into one, with a `feature:` prefix? (sorry I recently fixed the docs which said `feat`, but I've always used `feature` ;-))

110y: @toupeira 

> Let me know if you see another solution, otherwise I think we can merge this.

Sorry, I don't have any solution for workaround...
I think best way is implementing Registration with other parameters like `sector_identifier_uri` in the future.

>  Can you squash the commits into one, with a feature: prefix?

Squashed! 32f688b

toupeira:
@110y cool, thanks again for your contribution!

PR #36, opened 2017-07-07
Pass Access Tokens to claim builders.

gigr:
There are specific cases where we need to customize the claims returned from our `UserInfo` endpoint based on the access token, which contains information about the user, application, grant, and more.

This PR passed the `access_token` as a third argument to the claims builder.

@toupeira

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ Doorkeeper::OpenidConnect.configure do
       "#{resource_owner.first_name} #{resource_owner.last_name}"
     end
 
-    claim :preferred_username, scope: :openid do |resource_owner, application_scopes|
+    claim :preferred_username, scope: :openid do |resource_owner, application_scopes, access_token|
       # Pass the resource_owner's preferred_username if the application has
       # `profile` scope access. Otherwise, provide a more generic alternative.
       application_scopes.exists?(:profile) ? resource_owner.preferred_username : "summer-sun-9449"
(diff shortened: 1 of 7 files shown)


toupeira:
@gigr thanks for the PR! Unfortunately #33 just introduced conflicting changes, please rebase and also [add a `feature:` prefix to the commit message](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/CONTRIBUTING.md#commit-message-conventions).

toupeira:
@gigr do you have time to look into this? Otherwise I'll try to resolve the conflicts myself, would like to do a new release soon :)

gigr:
@toupeira Conflicts resolved!

gigr:
Is the new release you want to push going to be backwards-incompatible? If so, I would recommend we change the signature of claim builders to:

```ruby
claim :name do |resource_owner, access_token| do
  # Do stuff...
end
```

Passing the `scopes` and `access_token` as we do now is redundant—you can always get the scopes for the access token from `access_token.scopes`. 

If not, I guess we leave it as-is! 💃

toupeira:
> If not, I guess we leave it as-is! 💃

Yeah the other changes are all backwards-compatible so I agree we should just leave it, though the change you suggest does make sense :)

Thanks for resolving the conflicts, I'll merge this and will prepare a new release shortly!

PR #40, opened 2017-07-26
feature: hook for extra data in id_token

aisensiy:
Learn from [knock](https://github.com/nsarno/knock) to add extra payload to `id_token`.

This may have some conflicts with claims but the claims are used in `/userinfo` access. And the payload for `id_token` make it more useful to communicate through multi sub systems. So I can get enough info about the user without access to `/userinfo` url.

### DIFF
--- a/lib/doorkeeper/openid_connect/id_token.rb
+++ b/lib/doorkeeper/openid_connect/id_token.rb
@@ -21,7 +21,7 @@ module Doorkeeper
           iat: issued_at,
           nonce: nonce,
           auth_time: auth_time,
-        }
+        }.merge(@resource_owner.try(:to_token_payload) || {})
       end
 
       def as_json(*_)
--- a/spec/lib/id_token_spec.rb
+++ b/spec/lib/id_token_spec.rb
@@ -30,6 +30,31 @@ describe Doorkeeper::OpenidConnect::IdToken do
     end
   end
 
+  describe '#claims with extra payload' do
+
+    it 'returns all default claims' do
+
+      def user.to_token_payload 
+        {
+          username: 'test_name'
+        }
+      end
+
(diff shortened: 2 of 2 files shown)


aisensiy:
@toupeira Any update?

toupeira:
Thanks for the PR @aisensiy, unfortunately I currently don't have time to look into this but hope to do so in a couple of weeks!

Note that [the spec defines the `claims` parameter](http://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter) to specify which claims should be returned from userinfo and which should be included in the ID token, but adding support for that would be a bit more involved.

aisensiy:
The spec make me confusing... If the `scope` declares which `claim` to show in `userinfo` url then why not just embed the info into `id_token`?

toupeira:
@aisensiy see [section 5.4](http://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims):

> The Claims requested by the `profile`, `email`, `address`, and `phone` scope values are returned from the UserInfo Endpoint, as described in Section 5.3.2, when a `response_type` value is used that results in an Access Token being issued. However, when no Access Token is issued (which is the case for the `response_type` value `id_token`), the resulting Claims are returned in the ID Token.

toupeira:
I'm not sure this PR is a good idea, if you need this in your app I think it's better to just monkey-patch it until we support a proper way of doing this.

aisensiy:
So if the `response_type` is `id_token` all the info should be returned? But I am not sure that current architecture is good for extend this feature.

toupeira:
@aisensiy yeah if I understand correctly, but we currently don't support `response_type=id_token` ([also see this issue](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/42))

PR #44, opened 2017-12-27
Implicit flow

nhance:
Adds support for a grant type of `implicit_oidc` that will allow openID connect clients to connect.

Includes numerous changes from @stevenvegt

Probably not quite ready for review, but I don't know when I'll have a chance to pick it up and finish the tests. This is a great starting point for anyone who wants to add the required OIDC token support

PR #45, opened 2018-02-07
Implicit flow (`response_type=id_token` and `response_type=id_token token`)

nashby:
That's improved version of https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/44

I added support both for `response_type=id_token` and `response_type=id_token token`. Also it includes specs.

Also I send a PR to fix this https://github.com/nashby/doorkeeper-openid_connect/blob/966dd1ff412b46ec1e6075ea2d335bbe764d203a/lib/doorkeeper/openid_connect/request_extension.rb in Doorkeeper itself (https://github.com/doorkeeper-gem/doorkeeper/pull/1028) so we'll be able to remove this hack when fix is released.


doorkeeper-openid_connect.gemspec
```diff
@@ -25,6 +25,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rspec-rails'
   spec.add_development_dependency 'factory_girl'
   spec.add_development_dependency 'sqlite3'
-  spec.add_development_dependency 'pry-byebug'
+  spec.add_development_dependency 'pry-byebug', '3.4.3'
+  spec.add_development_dependency 'byebug', '~> 9.0'
```
toupeira: Please either remove the version constraints, or add a comment explaining why they're necessary. Also `byebug` is already required by `pry-byebug` so maybe we can remove it?

lib/doorkeeper/openid_connect/request_extension.rb
```diff
@@ -0,0 +1,16 @@
...
+           Doorkeeper::Request::IdTokenToken
+        else
+          get_strategy_without_multitype(grant_or_request_type, available)
+        end
+      end
+      alias_method :get_strategy_without_multitype, :get_strategy
+      alias_method :get_strategy, :get_strategy_with_multitype
+    end
+  end
+end
```
toupeira: I think this can be removed now with https://github.com/doorkeeper-gem/doorkeeper/pull/1028 merged?

lib/doorkeeper/openid_connect/response_types_config.rb
```diff
@@ -0,0 +1,17 @@
+module Doorkeeper
+  module OpenidConnect
+    module ResponseTypeConfig
+      private def calculate_authorization_response_types
+        types = super
+        if grant_flows.include? 'implicit_oidc'
```
toupeira: This probably requires some explanation in the README, I assume this needs to be enabled with `grant_flows` in `config/initializers/doorkeeper.rb`?

toupeira:
@nashby this looks excellent, thanks for the PR!

Before merging I'd like to test this with the [OIC conformance testing app](https://op.certification.openid.net:60000/), I should probably get around to it next week.

If you could add some docs to the README in the meantime that would be great!

lib/doorkeeper/openid_connect/request_extension.rb
nashby: if we remove it, this gem will only be compatible with Doorkeeper 4.3.0 (latest version). If that's ok I can remove it.

nashby:
@toupeira updated!

lib/doorkeeper/openid_connect/request_extension.rb
nashby: Actually I decided that it makes sense to drop this code and support only new version of Doorkeeper so we don't need to support legacy stuff.

toupeira: @nashby yeah I agree, thanks!

nashby:
@toupeira hey, do you need any help with testing this feature against https://op.certification.openid.net:60000/? I could probably take a look

toupeira:
@nashby thanks but it's alright :) So I had an old Heroku app lying around and just upgraded it to use your branch, it's deployed now at https://doorkeeper-connect-provider.herokuapp.com/ and the certification instance is up at https://op.certification.openid.net:60734/. The `response_type` can be changed at https://op.certification.openid.net:60734/pedit

The most straightforward test is the second one from the top ("OP-Response-code-*"), I just ran that now with all the different response types and both `id_token` and `id_token+token` are working successfully :+1:  With the Hybrid ones we get the error message that the response type isn't supported so that's all good too. Feel free to click around some more if you like, the tests are quite extensive ;-)

I'd say we can safely merge this now, thanks again for the solid work! I'll release a new version tomorrow or next week at the latest. I'll also be happy to add you as a maintainer as you asked, and I'll see that I can clean up the example app a bit and push it to a public repository.

toupeira:
@nashby almost forgot: can you squash down the commits into one? Thanks :)

nashby:
@toupeira done!

toupeira:
@nashby big thanks! Just released 1.3.0 with this :rocket:

nashby:
@toupeira awesome, thanks!

PR #47, opened 2018-03-13
Redirect errors per Section 3.1.2.6 of OpenID Connect 1.0

ryands:
Error handling currently returns a json object whose body contains the error code and description along with an HTTP status 401.  This pull-request changes the behavior of error handling in the gem to conform with [OpenID Connect 1.0 Section 3.1.2.6](http://openid.net/specs/openid-connect-core-1_0.html#AuthError):

> Unless the Redirection URI is invalid, the Authorization Server returns the
> Client to the Redirection URI specified in the Authorization Request with the
> appropriate error and state parameters. Other parameters SHOULD NOT be
> returned.


toupeira:
@ryands thanks for your PR! :+1:

PR #49, opened 2018-05-04
Set id_token when nil (RefreshToken)

Miouge1:
This adds the id_token as part of the refresh token response as permitted by the OIDC spec:
> the response body is the Token Response of Section 3.1.3.3 except that it might not contain an id_token

Many OIDC providers, like Google, have that behavior so it's used in things like Kubernetes:
> Providers that don’t return an id_token as part of their refresh token response (e.g. Okta) aren’t supported by this plugin and should use “Option 2” below.
[source](https://kubernetes.io/docs/admin/authentication/#option-1---oidc-authenticator)

See also https://gitlab.com/gitlab-org/gitlab-ce/issues/44485

### DIFF
--- a/lib/doorkeeper/openid_connect/oauth/token_response.rb
+++ b/lib/doorkeeper/openid_connect/oauth/token_response.rb
@@ -7,7 +7,7 @@ module Doorkeeper
         def body
           if token.includes_scope? 'openid'
             super
-              .merge(id_token: id_token.try(:as_jws_token))
+              .merge(id_token:  id_token ? id_token.as_jws_token : Doorkeeper::OpenidConnect::IdToken.new(token).as_jws_token)
               .reject { |_, value| value.blank? }
           else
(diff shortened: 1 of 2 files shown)


Miouge1:
@nashby, OK I'll see what I can do.

Miouge1:
@nashby tests are added. Travis is not happy on Ruby 2.2 but I don't think that it's related to this change.

nashby:
@Miouge1 thank you!

Miouge1:
Cool! Anything I can do to make that part of the 1.4.0 release?

nashby:
@Miouge1 I think that's up to @toupeira

toupeira:
@Miouge1 thanks for your PR! :+1: 

1.4.0 is now released including your change, as well as dropping support for Ruby < 2.3 to fix the CI issues.

PR #51, opened 2018-06-07
Update deprecated `jws_private_key` config item

RobinDaugherty:
`jws_private_key` [is deprecated](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/cdabae87f814877b1c4631a638f0ba6ca3af334d/lib/doorkeeper/openid_connect/config.rb#L31) and the initializer template was updated to reflect that but the readme wasn't updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,8 +115,8 @@ The following settings are required in `config/initializers/doorkeeper_openid_co
     end
     ```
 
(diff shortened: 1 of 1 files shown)


nashby:
@RobinDaugherty good catch! Thanks!

Issue #16, opened 2015-05-21
Claims in IdToken


gilbertquri:
Why does IdToken not incorporate user defined claims? Is it because you want to minimize the token size? Perhaps it should respect the scopes parameter and incorporate user defined claims if a certain scope (e.g. profile) is present in the scopes?

toupeira:
The spec has this to say:

http://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims
> The Claims requested by the `profile`, `email`, `address`, and `phone` scope values are returned from the UserInfo Endpoint, as described in Section 5.3.2, when a `response_type` value is used that results in an Access Token being issued. However, when no Access Token is issued (which is the case for the `response_type` value `id_token`), the resulting Claims are returned in the ID Token. 

This is currently not implemented, i.e. additional claims are always returned from the UserInfo endpoint. We also don't support `response_type=id_token` yet, or the `claims` parameter which would allow the client to specify where each claim should be returned (see http://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter).

I don't have any plans to implement this yet, but I want to improve the UserInfo endpoint so it filters any Standard Claims that aren't covered by the requested scopes.

aisensiy:
@gilbertquri 

[https://github.com/aisensiy/doorkeeper-openid_connect](https://github.com/aisensiy/doorkeeper-openid_connect) I have a fork and implement a way to extend the claims in id token.

toupeira:
See related discussion in https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/40

toupeira:
Also note that support for `response_type=id_token` was added recently in https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/45

PR #52, opened 2018-06-26
Support custom claims in ID token

toupeira:
Closes https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/16

The current behaviour of only returning custom claims from `/user_info` was kept to avoid unintended leakage / growth of the ID token.


toupeira:
@nashby I could use some eyes on this :)

@aisensiy and @gilbertquri please also take a look, let me know if this is missing someting for your use-cases.

nashby:
@toupeira Nice work, looks good to me!

PR #55, opened 2018-08-02
Loose doorkeeper version.

### LINKED ISSUE #54: Compatibility with Doorkeeper 5.0.0rc1?
Hi, thanks for the great work on this gem.  I'm trying to gauge how compatible a recent version of this gem might be with Doorkeeper 5.0.0rc1.  I know there's probably not good support for it since it's an RC, but I wanted to ask what options I might have.  I'm pinned to rc1 because I need PKCE for my project.

Thanks!

nashby:
closes #54 

@toupeira hey, I tested my production app with latest Doorkeeper release candidate and all my tests are green so I think we can let people use it with Doorkeeper OpenidConnect.

### DIFF
--- a/doorkeeper-openid_connect.gemspec
+++ b/doorkeeper-openid_connect.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 2.3"
 
-  spec.add_runtime_dependency 'doorkeeper', '~> 4.3'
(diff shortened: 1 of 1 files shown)


toupeira:
@nashby sorry for the delay! Had a look through https://github.com/doorkeeper-gem/doorkeeper/blob/master/NEWS.md and there don't seem to be any changes that affect us, so I'll merge this and release a new version.

Issue #63, opened 2019-02-07
When `prompt=none` is set, reauthorizing with a smaller set of scopes causes an error


meagar:
If I have previously authorized with a Doorkeeper-OIDC-based IDP using `scope=openid+email+name`, an authorization request with a *more narrow* set of claims, such as `scope=openid+email` will fail. The error will indicate that user consent is required, however I think this is incorrect. The user has already granted consent for the `email` and `name` claims; my reauthorization request is only requesting the `email` claim, so I believe the correct behavior is to generate an return a new access token with the `openid` and `email` claims.

If this is in fact a bug, the problem seems to be that an *exact match* between the incoming scopes and previously granted access token's scopes is required by this method:

```
def matching_tokens_for_resource_owner(owner)
  Doorkeeper::AccessToken.authorized_tokens_for(pre_auth.client.id, owner.id).select do |token|
    Doorkeeper::AccessToken.scopes_match?(token.scopes, pre_auth.scopes, pre_auth.client.scopes)
  end
end
```

toupeira:
@meagar thanks for the report! I think you're right:

> The user has already granted consent for the `email` and `name` claims; my reauthorization request is only requesting the `email` claim, so I believe the correct behavior is to generate and return a new access token with the `openid` and `email` claims.

That is what I would expect as well, I can't find anything concrete in the specs except for [this section](https://tools.ietf.org/html/rfc6749#section-1.5):

> Refresh tokens are issued to the client by the authorization server and are used to obtain a new access token when the current access token becomes invalid or expires, **or to obtain additional access tokens with identical or narrower scope**

I've started to look into this in https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/65, but the problem is that Doorkeeper also performs an exact match in https://github.com/doorkeeper-gem/doorkeeper/blob/9ae5f9fff7234123216ae56f37c73534a7373654/lib/doorkeeper/models/access_token_mixin.rb#L104. So even though the OIDC code won't raise an error anymore, Doorkeeper will still render the authorization form.

I'll open an issue on the Doorkeeper gem!

PR #65, opened 2019-02-08
WIP: Support reauthorization with a narrower set of scopes

toupeira:
Fixes https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/63


phlegx:
News from this PR?

toupeira:
@phlegx sorry, no updates since I last worked on this!

From what I remember, this would also require some changes in Doorkeeper itself. IIRC one part is changing the token lookup, but another (trickier) part is also issuing a new token with the reduced set of scopes.

Issue #61, opened 2019-02-07
Open redirect vulnerability when `prompt=none`


meagar:
tl;dr  Doorkeeper OIDC can be provoked to blindly redirect the browser to whatever site is sent in `redirect_uri` paramter when `prompt=none`.

---

Given an otherwise valid authorization attempt with `prompt=none`, if an error can be provoked, the user is redirected to whatever redirect URI is passed in through the `redirect_uri` parameter with an error message in the query string.

For example, assume I have a Doorkeeper-backed Identity provider at `identity-rovider.com`. It has an application with a client ID of `123` and a redirect-URI of `https://real-site.com/callback`.

If I have previously authorized with `identity-provider.com` with the `openid profile` claims, and I issue a new authorization request with only the `openid` scope, and `prompt=none`, an error is raised because the claims I'm requesting don't match the claims I previously authorized, but `prompt=none` prevents me from reauthorizing.

The error causes me to be redirected to whatever `redirect_uri` is included in the request, **even if it's not a valid whitelisted redirect URI for the subject application**.

A sample request, assuming I'm already authorized (newlines added for clarity):

    https://identity-provider.com/oauth/authorize?
      client_id=123
      redirect_uri=https%3A%2F%2Fattacksite.com&
      response_type=code&
      scope=openid&
      prompt=none

This request sends me to (again, newlines added for clarity):

    https://attacksite.com/?
      error=consent_required&
      error_description=The+authorization+server+requires+end-user+consent

At this point, `attacksite.com` can display a phishing form; given that I clicked on a link that started with `identity-provider.com` to start the authorization flow, it's possible that I can be tricked into entering my credentials into the attacking site. This is (I believe) an example of a [Dangerous URL Redirect](https://www.owasp.org/index.php/Unvalidated_Redirects_and_Forwards_Cheat_Sheet#Dangerous_URL_Redirects).

The problem seems to be in how exceptions are rescued:

https://github.com/meagar/doorkeeper-openid_connect/blob/12685b19af46e0e1abcf6c74efe919268fec34f8/lib/doorkeeper/openid_connect/helpers/controller.rb#L22

```
error = ::Doorkeeper::OAuth::ErrorResponse.new(
  name: exception.error_name,
  state: params[:state],
  redirect_uri: params[:redirect_uri])
```

This line extracts `redirect_uri` directly from the incoming parameters, and does nothing to validate it before redirecting the browser there on subsequent lines.

[...]

PR #66, opened 2019-02-08
Validate redirect_uri when handling custom params

toupeira:
Fixes https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/61

### DIFF
--- a/spec/controllers/doorkeeper/authorizations_controller_spec.rb
+++ b/spec/controllers/doorkeeper/authorizations_controller_spec.rb
@@ -65,20 +65,7 @@ describe Doorkeeper::AuthorizationsController, type: :controller do
           expect_successful_callback!
         end
 
-        it 'renders a login_required error when not logged in' do
-          authorize! prompt: 'none', current_user: nil, state: 'somestate'
-
-          error_params = {
-            'error' => 'login_required',
-            'error_description' => 'The authorization server requires end-user authentication',
-            'state' => 'somestate'
-          }
-
-          expect(response).to redirect_to build_redirect_uri(error_params)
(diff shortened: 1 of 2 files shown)


spec/controllers/doorkeeper/authorizations_controller_spec.rb
```diff
@@ -89,10 +76,35 @@ def expect_successful_callback!
...
+
+            expect(response).to redirect_to build_redirect_uri(error_params)
+            expect(JSON.parse(response.body)).to eq(error_params)
+          end
+
+          it 'does not redirect to an invalid redirect_uri' do
+            authorize! prompt: 'none', current_user: nil, state: 'somestate', redirect_uri: 'https://evilapp.com'
+
+            expect(response).not_to be_redirect
+            expect(response.status).to eq 401
```
toupeira: See discussion in https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/61#issuecomment-461883958:

> > @meagar: I think the correct behaviour here is to render a 422 error, rather than redirect the browser, regardless of the `prompt=none`.
>
> @toupeira: I'm not sure if the 422 error should have precedence, since we're already returning the OIDC/OAuth error.

toupeira: I've now changed the code so it checks `pre_auth.valid?` and prioritizes Doorkeeper's OAuth error response before falling back to the OIDC error.

This is currently also a 401, but see https://github.com/doorkeeper-gem/doorkeeper/issues/550 about maybe changing this to a 400 as suggested by the OAuth specs.

PR #69, opened 2019-02-21
Add 'crv' param to normalized EC signing key

### LINKED ISSUE #68: The Elliptic Curve's `crv` parameter is missing from the discovery/keys JSON response
For my OP (_OpenID Provider_) Rails application, I generated just one key-pair with the following commands:

```bash
openssl ecparam -genkey -name prime256v1 -noout -out priv-key.pem
openssl ec -in priv-key.pem -pubout -out pub-key.pem
```

Then I added the Doorkeeper-OpenID-connect routes (`use_doorkeeper_openid_connect`), which of course includes the OpenID Connect Discovery paths.

Then, after setting everything up properly in order to use the key-pair above, I perform a GET request to the correct Discovery `jwks_uri` (e.g. `https://example.com/oauth/discovery/keys`) to get the signing key(s) that an RP (_Relying Party_) should use to validate the signatures.

### Expected response

The output of my request to the `jwks_uri` should be a JSON-serialized response, structured as a valid _JSON Web Key_ (_JWK_ &rarr; [RFC-7517](https://tools.ietf.org/html/rfc7517)) _Set_ containing a single key, whose cryptographic algorithm's properties should be structured according to the _JSON Web Algorithms_ (_JWA_ &rarr; [RFC-7518](https://tools.ietf.org/html/rfc7518)) specification. Therefore, it should look more or less like this:

```json
{
  "keys": [ 
    {
      "kty": "EC",
      "kid": "XsXAbPnb1Z25OCWBdrW8112nIuTDfQ8sITTCqELEtbo",
      "x": "7pp2uTZUXR1Vq1EYBPqvpHSt7_TqtBZPVl8nrpLAOpc",
      "y": "IMKuZb2fr7BqnKaHDVIELZJV8RtX1JtN77ZD3fSiO5g",
      "use": "sig",
      "alg": "ES256",
      "crv": "P-256"
    }
  ]
}
```

### Actual response

This is the actual response:

```json
{
  "keys": [ 
    {
      "kty": "EC",
      "kid": "XsXAbPnb1Z25OCWBdrW8112nIuTDfQ8sITTCqELEtbo",
      "x": "7pp2uTZUXR1Vq1EYBPqvpHSt7_TqtBZPVl8nrpLAOpc",
      "y": "IMKuZb2fr7BqnKaHDVIELZJV8RtX1JtN77ZD3fSiO5g",
      "use": "sig",
      "alg": "ES256"
    }
  ]
}
```

As you can see, it is not that different from what I expected, with the important exception that the `crv` property is missing.

According to the _JWA_ specs, [section 5.2.1](https://tools.ietf.org/html/rfc7518#section-6.2.1), `crv` is a member that _"MUST be present for all Elliptic Curve public keys"_.

Going further, a short example can be made to highlight the possible bad consequences. Let's suppose to write an RP app which first of all acquires the _JWK-Set_ from the Discovery service of the OP. Then, after a certain event (whose details are irrelevant for this example), the OP sends to the RP a _JSON Web Token_ (_JWT_) message, signed according to _JSON Web Signature_ (_JWS_). The RP app starts decoding and validating the _JWS_ string, for example using the [`json-jwt` Ruby Gem](https://github.com/nov/json-jwt), thus invoking the method `JSON::JWT.decode`, passing to it the raw _JWT_/_JWS_ string and the _JWK-Set_ data acquired earlier. As a result, the process of decoding and validation fails due to the missing _EC Curve_ (`crv`) information. In the case of _json-jwt_ Gem, you would get an `UnknownAlgorithm` error, with message `Unknown EC Curve`.

Note: I consider the `json-jwt` Gem being a good example, since it's used in turn by this very Gem.

---

If you believe this is an actual bug - I mean, a missed specification compliance - I should be able to submit soon a possible fix (unless somebody precedes me, or discards the issue).

marco-nicola:
This change fixes the DiscoveryController#keys JSON response
not being compliant with RFC 7518 (JWA), when representing the
parameters for Elliptic Curve Keys.

See: https://tools.ietf.org/html/rfc7518#section-6.2.1

Closes #68

### DIFF
--- a/lib/doorkeeper/openid_connect.rb
+++ b/lib/doorkeeper/openid_connect.rb
@@ -56,7 +56,7 @@ module Doorkeeper
       when :RSA
         key.slice(:kty, :kid, :e, :n)
       when :EC
-        key.slice(:kty, :kid, :x, :y)
+        key.slice(:kty, :kid, :crv, :x, :y)
       when :oct
         key.slice(:kty, :kid)
       end
--- a/spec/controllers/discovery_controller_spec.rb
+++ b/spec/controllers/discovery_controller_spec.rb
@@ -108,7 +108,7 @@ describe Doorkeeper::OpenidConnect::DiscoveryController, type: :controller do
     context 'when using an EC key' do
       before { configure_ec }
 
-      it_behaves_like 'a key response', expected_parameters: [:kty, :kid, :x, :y, :use, :alg]
+      it_behaves_like 'a key response', expected_parameters: [:kty, :kid, :crv, :x, :y, :use, :alg]
     end
 
     context 'when using an HMAC key' do
--- a/spec/lib/openid_connect_spec.rb
+++ b/spec/lib/openid_connect_spec.rb
@@ -33,6 +33,7 @@ describe Doorkeeper::OpenidConnect do
         expect(subject.signing_key_normalized).to eq(
           'kty' => :EC,
           'kid' => 'dOx_AhaepicN2r2M-sxZhgkYZMCX7dYhPsNOw1ZiFnI',
+          'crv' => :'P-521',
           'x' => 'AeYVvbl3zZcFCdE-0msqOowYODjzeXAhjsZKhdNjGlDREvko3UFOw6S43g-s8bvVBmBz3fCodEzFRYQqJVI4UFvF',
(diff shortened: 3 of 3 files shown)


toupeira:
@marco-nicola thanks for the PR and detailed issue! :grinning: 

The spec seems pretty clear on this so I'll merge it.

marco-nicola:
Thank you! My pleasure :relaxed:

PR #70, opened 2019-03-05
fix runtime dependency

michaelglass:
since #67, doorkeeper-open-id_connect has called an API (`#mapping`) that was introduced in doorkeeper 5.0. (https://github.com/doorkeeper-gem/doorkeeper/pull/1082/files#diff-30d1c02eb523d3fdbe0ea87cd811bc3fR7)

See more here: https://github.com/doorkeeper-gem/doorkeeper-openid_connect/commit/3893ea1318e302a078272ff408aff1c50bef85d1#r32599682


toupeira:
@michaelglass thanks for catching this! :)

I'll release a new version shortly.

PR #72, opened 2019-04-15
Add revocation_endpoint and introspection_endpoint to .well-known

scarfacedeb:
This PR adds `revocation_endpoint` and `introspection_endpoint` keys to `/.well-known/openid-configuration` response.

Although non-standard, there're multiple examples that provide those keys:

- https://accounts.google.com/.well-known/openid-configuration
- https://developer.okta.com/docs/api/resources/oidc/#response-properties-7
- https://connect2id.com/products/server/docs/api/discovery#openid-config
- https://developers.onelogin.com/openid-connect/api/provider-config

Taking into the consideration that doorkeeper already provides those endpoints, I think it makes sense to mention them in the discovery response as well.


toupeira:
@scarfacedeb sorry for the late response, and thanks for your contribution! :heart: 

I agree this makes sense, merging...

Issue #71, opened 2019-03-11
Grant Type Password Requires a client_id


jeffsawatzky:
As per https://github.com/doorkeeper-gem/doorkeeper/pull/296, client_id is optional for the password grant type, however doorkeeper-openid_connect seems to assume that it will always be there.

toupeira:
@niltz thanks, can you give some more details on the errors you're seeing? I'm not sure if this is related to [this change](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/commit/3893ea1318e302a078272ff408aff1c50bef85d1#diff-075af0c695b538d6def8b65a98eb3cc8L10), as it seems that's the only file where we deal with `client`.

natashad:
@toupeira The issue seems to be in the code to get the `audience` for the id token. https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/lib/doorkeeper/openid_connect/id_token.rb#L49. If there's no client_id passed in, there's no application so `@access_token.application.uid` this fails with a `NoMethodError (undefined method 'uid' for nil:NilClass)`

```
FATAL
FATAL   NoMethodError (undefined method `uid' for nil:NilClass):
FATAL
FATAL   doorkeeper-openid_connect (1.6.0) lib/doorkeeper/openid_connect/id_token.rb:49:in `audience'
doorkeeper-openid_connect (1.6.0) lib/doorkeeper/openid_connect/id_token.rb:19:in `claims'
doorkeeper-openid_connect (1.6.0) lib/doorkeeper/openid_connect/id_token.rb:28:in `as_json'
doorkeeper-openid_connect (1.6.0) lib/doorkeeper/openid_connect/id_token.rb:32:in `as_jws_token'
doorkeeper-openid_connect (1.6.0) lib/doorkeeper/openid_connect/oauth/token_response.rb:12:in `body'
doorkeeper (5.0.0) app/controllers/doorkeeper/tokens_controller.rb:6:in `create'
app/controllers/tokens_controller.rb:11:in `create'
actionpack (5.2.2) lib/abstract_controller/base.rb:194:in `process_action'
actionpack (5.2.2) lib/action_controller/metal/instrumentation.rb:34:in `block in process_action'
```

alduro:
I'm running exactly into this issue. Is this still being maintained ?

toupeira:
[...]

PR #77, opened 2019-07-16
Prevent 500 when required parameters are not present

cincospenguinos:
Currently, the `pre_auth` object is getting inspected in `authenticate_resource_owner!`, before Doorkeeper validates it.  If the required OAuth params are not present, it will currently blow up with `undefined method application for nil`. Our pull request ensures a `401` error is returned rather than a `500` error for this bit.

Solves https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/78
This appears to be related to #71.

### DIFF
--- a/lib/doorkeeper/openid_connect/helpers/controller.rb
+++ b/lib/doorkeeper/openid_connect/helpers/controller.rb
@@ -8,6 +8,7 @@ module Doorkeeper
           super.tap do |owner|
             next unless controller_path == Doorkeeper::Rails::Routes.mapping[:authorizations][:controllers] &&
               action_name == 'new'
+            raise Errors::OpenidConnectError unless pre_auth.valid?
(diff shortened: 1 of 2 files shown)


cincospenguinos:
See #79

lib/doorkeeper/openid_connect/helpers/controller.rb
```diff
@@ -8,6 +8,7 @@ def authenticate_resource_owner!
           super.tap do |owner|
             next unless controller_path == Doorkeeper::Rails::Routes.mapping[:authorizations][:controllers] &&
               action_name == 'new'
+            raise Errors::OpenidConnectError unless pre_auth.valid?
```
urnf: This will still fail on line 37 if we're missing a `redirect_uri` in our error_response

Issue #78, opened 2019-07-16
Not passing required parameters results in a 500


isabellechalhoub:
The `Doorkeeper::AuthorizationsController` checks to see whether or not required parameters are present by running `pre_auth.authorizable?` [in the `new` action](https://github.com/doorkeeper-gem/doorkeeper/blob/03b1437c6bda10ec74c256b4f8c5c3d98b4e5227/app/controllers/doorkeeper/authorizations_controller.rb#L7).

This gem modifies the behavior of the `:authenticate_resource_owner!` before filter that runs prior to the `new` action.  One of the modifications is to inspect the `pre_auth` object (an instance of `OAuth::PreAuthorization`) and [look at its `scopes`](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/17dbe9517e855ed46d6b8dec3a311e093933e34e/lib/doorkeeper/openid_connect/helpers/controller.rb#L11).

The `OAuth::PreAuthorization#scopes` method in turn fires the [`build_scopes` private method](https://github.com/doorkeeper-gem/doorkeeper/blob/03b1437c6bda10ec74c256b4f8c5c3d98b4e5227/lib/doorkeeper/oauth/pre_authorization.rb#L59), which calls `client.application.scopes`.  This method assumes that `client` is present.  However, if you do not initialize the object with `@client`, it obviously blows up.  This has not been an issue in Doorkeeper itself, because usually the `OAuth::PreAuthorization` object is assumed to be valid.  However, this OpenID Connect gem inspects the object before it has been validated.

We have a PR to add a validation check before trying to inspect the scopes.

toupeira:
Thanks again, you can give this a try with the new version 1.6.2.

PR #79, opened 2019-07-25
Defer to Doorkeeper when scopes are missing

cincospenguinos:
As we explored our use case, we realized that we did not want to submit a 4xx error on missing parameters for openid connect within our Oauth flow. We refactored a bit and created a test case that proved that, if parameters were missing, doorkeeper-openid_connect would pass on its opportunity to interact with the flow and doorkeeper would take over.

The biggest change to pay attention to is the extraction of `pre_auth.scopes` out to a separate private method. This method captures a `NoMethodError` that gets thrown when pre_auth attempts to retrieve the scopes of an object that has no client, and therefore no application. This refactor is really a workaround--lacking the context concerning this flow, I decided that simply capturing the error and navigating around it would be the best way to go.

If any of you have any suggestions or desires to see this area move in a different direction, let me know and my team and I will adjust our PR to reflect that.

Solves #78
This appears to be related to #71.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,5 +291,3 @@ Doorkeeper::OpenidConnect is released under the [MIT License](http://www.opensou
 ## Sponsors
 
 Initial development of this project was sponsored by [PlayOn! Sports](https://github.com/playon).
-
-Can I do this?
--- a/README.md
+++ b/README.md
@@ -291,3 +291,5 @@ Doorkeeper::OpenidConnect is released under the [MIT License](http://www.opensou
 ## Sponsors
 
 Initial development of this project was sponsored by [PlayOn! Sports](https://github.com/playon).
+
+Can I do this?


lib/doorkeeper/openid_connect/helpers/controller.rb
```diff
@@ -6,21 +6,34 @@ module Controller
...
+          handle_error(exception)
+        end
+
+        def valid_controller_path?
+          controller_path == Doorkeeper::Rails::Routes.mapping[:authorizations][:controllers] && action_name == 'new'
+        end
+
+        def scopes
+          pre_auth.scopes
+        rescue NoMethodError
```
urnf: To be clear - this PR is obviously meant to raise questions and NOT to be merged - we're curious if we need to guarantee the validity of `pre_auth` after `resource_owner_authenticator` is run.  Are we supposed to verify that `pre_auth` is valid with a valid `client` before we get to this point?

If not, calling `scopes` can attempt to `build_scopes` in the PreAuthorization with a `nil` client.

urnf: Thanks for looking at this @toupeira !

I might be missing something about this, so correct me if I'm wrong with my assumptions:
- `Doorkeeper::OpenIDConnect::Helpers::Controller` overrides base Doorkeeper's `authenticate_resource_owner!`, calls it (with `super`), opens the resource owner returned, and runs OIDC specific behavior (`prompt` analysis + `max_age` check)
- In order to do so, it currently requires that `pre_auth.client` is valid.  Every other part of the PreAuthorization can be missing - `redirect_uri`, `scopes`, and `response_type` can be `nil` and it'd still work fine because only `scopes` is used in this context and `scopes` in PreAuthorization does a `client` lookup if it's missing.
- Doorkeeper's `authenticate_resource_owner!` is calling doorkeeper config  `resource_owner_authenticator`

Because of that:
- Would our `resource_owner_authenticator` need to validate `pre_auth.client?` so that by the time we drop into `Doorkeeper::OpenIDConnect::Helpers::Controller:authenticate_resource_owner!` we guarantee that `pre_auth.scopes` doesn't raise an exception trying to grab `client` if `scopes` doesn't exist.

I noticed that https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/80 puts it back in, so I think that answers the question.

That being said, I'm wondering if it's worth checking `pre_auth.valid?` rather than specifically `pre_auth.client`.  

Would it make sense to ensure that `response_type`, `client`, `redirect_uri` are valid before we take action (they seem required in the OIDC RFC?) or is it something you want to leave to the base Doorkeeper gem itself when we finish and it drops down to the `authorized_applications_controller` or `authorizations_controller`?

Let me know if it's a concern I should put on that PR.  Thanks!

PR #80, opened 2019-08-08
Check for client presence in controller

toupeira:
This should fix https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/78, by adding back the check for `pre_auth.client` which was removed in https://github.com/doorkeeper-gem/doorkeeper-openid_connect/commit/3893ea1

Also, refactor `Doorkeeper::OpenidConnect::Helpers::Controller` a bit:
- include `oidc` in all custom method names to protect against name clashes with the `doorkeeper` gem
- split out `oidc_authorization_request?` and `handle_oidc_error!`

Thanks to @cincospenguinos @urnf @isabellechalhoub for the inspiration! :wink:

### DIFF
--- a/lib/doorkeeper/openid_connect/helpers/controller.rb
+++ b/lib/doorkeeper/openid_connect/helpers/controller.rb
@@ -6,14 +6,23 @@ module Doorkeeper
 
         def authenticate_resource_owner!
           super.tap do |owner|
-            next unless controller_path == Doorkeeper::Rails::Routes.mapping[:authorizations][:controllers] &&
-              action_name == 'new'
-            next unless pre_auth.scopes.include?('openid')
+            next unless oidc_authorization_request?
 
-            handle_prompt_param!(owner)
-            handle_max_age_param!(owner)
+            handle_oidc_prompt_param!(owner)
+            handle_oidc_max_age_param!(owner)
           end
         rescue Errors::OpenidConnectError => exception
+          handle_oidc_error!(exception)
+        end
+
+        def oidc_authorization_request?
+          controller_path == Doorkeeper::Rails::Routes.mapping[:authorizations][:controllers] &&
+            action_name == 'new' &&
+            pre_auth.client &&
+            pre_auth.scopes.include?('openid')
+        end
+
+        def handle_oidc_error!(exception)
           # clear the previous response body to avoid a DoubleRenderError
           self.response_body = nil
 
@@ -39,7 +48,7 @@ module Doorkeeper
           end
         end
 
-        def handle_prompt_param!(owner)
+        def handle_oidc_prompt_param!(owner)
           prompt_values ||= params[:prompt].to_s.split(/ +/).uniq
 
           prompt_values.each do |prompt|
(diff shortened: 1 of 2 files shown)


toupeira:
@cincospenguinos @urnf @isabellechalhoub if you could give this a quick review that would be great!

lib/doorkeeper/openid_connect/helpers/controller.rb
```diff
@@ -6,14 +6,23 @@ module Controller
...
+            handle_oidc_max_age_param!(owner)
           end
         rescue Errors::OpenidConnectError => exception
+          handle_oidc_error!(exception)
+        end
+
+        def oidc_authorization_request?
+          controller_path == Doorkeeper::Rails::Routes.mapping[:authorizations][:controllers] &&
+            action_name == 'new' &&
+            pre_auth.client &&
```
urnf: Thanks!  This should fix our issue.

Just mirroring a question I had here https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/79#discussion_r312297957 - would just checking `params[:scopes]` for 'openid' first or straight up verifying `pre_auth.valid?` work?  So something like this:

```ruby
controller_path == Doorkeeper::Rails::Routes.mapping[:authorizations][:controllers] &&
            action_name == 'new' &&
            params[:scope] &&
            params[:scope].include?('openid') &&
            pre_auth.valid?
```
or 
```ruby
controller_path == Doorkeeper::Rails::Routes.mapping[:authorizations][:controllers] &&
            action_name == 'new' &&
            pre_auth.valid? &&
            pre_auth.scopes.include?('openid')
```

In this case, we'd drop down to base Doorkeeper behavior if scope is empty or doesn't have 'openid', regardless of the state that `pre_auth` is in (since `params` should always be available).  If the scope _is_ correct, then we validate all of `pre_auth` before we even get started.

I'm hesitant to suggest this, since it may now tie implementation details that should remain in Doorkeeper (validating all of `pre_auth`) with OIDC, but are we already doing that reaching into `pre_auth.scopes` and `pre_auth.client`?  It may also be validating unnecessary params (as you noted, we only strictly need `client` since `scopes` can infer off `client` - but could Doorkeeper change the `pre_auth.scopes` implementation?)

In addition, if we go down this route, would it need to `raise Errors::OpenidConnectError unless pre_auth.valid?` instead of `next` and letting Doorkeeper handle the invalid pre_auth?  Unsure about that, that might be raising where Doorkeeper should be handling the error.

Sorry for all the questions; this looks good, I'm just wondering for future work.  And let me know if I'm just going about this all wrong.

toupeira: @urnf ah thanks, that's actually a great point!

I think it would make sense to reuse `pre_auth.valid?`, especially since we already call it in the error handling code below, and it would have avoided the open redirect vulnerability fixed in https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/66/files!

So I tried out what happens when we replace `pre_auth.client` with `pre_auth.valid?`, but one problem is that some errors will now return a HTML response from the Doorkeeper gem, rather than the expected JSON response from the OIDC gem. So I'm not sure how to proceed here, we might need some changes in the Doorkeeper gem to force a JSON response.

But yeah, this whole class is due for some major refactoring :see_no_evil: The only reason this is implemented through hooking into `authenticate_resource_owner!` was because we need access to the controller scope for certain actions (rendering/redirecting), and we can't hook into `Doorkeeper::AuthorizationsController#new` directly because users can [specify their own custom controllers](https://github.com/doorkeeper-gem/doorkeeper/wiki/Customizing-routes#changing-controllers).

If you want to keep looking into this feel free to open an issue or PR, I'm afraid I won't have too much time for it myself, and also just generally avoid dealing with OAuth for too long to keep me sane :stuck_out_tongue:

PR #81, opened 2019-08-29
Allow silent authentication without user consent

jarosan:
This PR fixes an issue I've experienced when doing silent authentication together with
```
  skip_authorization do |resource_owner, client|
    true
  end
```
setting in `Doorkeeper.configure`. Silent authorization will pass `prompt=none` which will check if there are any existing tokens in `matching_tokens_for_oidc_resource_owner`. This check is not needed when `skip_authorization` is set to true.


toupeira:
@jarosan thanks for this PR, and please excuse the delay!

Looks good to me, will merge and release a new version shortly...

PR #83, opened 2019-09-16
Fix Doorkeeper 5.2 compatibility

nbudin:
Doorkeeper 5.2.0 seems to have changed the method signature for `OAuth::PreAuthorization#initialize`.  It previously took a `client` as a parameter, but now expects a `:client_id` key in `attrs` instead.  (This probably should have been considered a breaking change but doesn't seem to have been.)

Because of that, trying to log in when the server is using this gem with doorkeeper 5.2.0 gives a traceback ending in:

```
doorkeeper-openid_connect (1.6.2) lib/doorkeeper/openid_connect/oauth/pre_authorization.rb:8:in `initialize'
doorkeeper (5.2.0) lib/doorkeeper/oauth/pre_authorization.rb:20:in `initialize'
ArgumentError processing GraphQL query: wrong number of arguments (given 3, expected 1..2)
```

This patch attempts to fix it while still maintaining compatibility with older Doorkeeper releases.  To do that, it checks both the type of the `client` parameter and whether or not the super initializer has a parameter called `client`.  It's a little complicated but it seems to work.

### DIFF
--- a/lib/doorkeeper/openid_connect/oauth/pre_authorization.rb
+++ b/lib/doorkeeper/openid_connect/oauth/pre_authorization.rb
@@ -4,8 +4,19 @@ module Doorkeeper
       module PreAuthorization
         attr_reader :nonce
 
-        def initialize(server, client, attrs = {})
-          super
+        def initialize(server, client_or_attrs, attrs = {})
+          if client_or_attrs.is_a?(ActionController::Parameters)
+            attrs = client_or_attrs
+            client = nil
+          else
+            client = client_or_attrs
+          end
+
+          if OAuth::PreAuthorization.method(:initialize).parameters.include?([:req, :client])
+            super(server, client, attrs)
+          else
+            super(server, attrs)
+          end
           @nonce = attrs[:nonce]
         end
       end


toupeira:
@nbudin thanks for the report! :+1: 

Technically we're trespassing on internal Doorkeeper APIs, so I'm not sure if it can really be considered a "breaking change" :stuck_out_tongue: 

But I stumbled across some other breakage with Doorkeeper 5.2, so I think I'll just require that as a minimum for doorkeeper-openid_connect 1.7.0 (or 2.0.0? :thinking:) and started https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/85 to get Doorkeeper 5.2 running, and also possibly Rails 6.

In the meantime I also released 1.6.3 which specifically requires Doorkeeper < 5.2.

Let me know if that works for you! I'll close this PR.

nbudin:
@toupeira That seems reasonable, thanks!  I'm currently using my fork on the one project that needs it, and I can keep doing so until you release the next version.

nbulaj:
Hi @toupeira . Sorry for the closed issue, but it the main topic here

> Technically we're trespassing on internal Doorkeeper APIs, so I'm not sure if it can really be considered a "breaking change"

I'm open for the suggestions on how we can make the API open for `doorkeeper-openid_connect` to simplify your (and everyone else) life. If you have some ideas - you can propose a PR for main project and we'll continue here.

I'm thinking about a plugin system for Doorkeeper for a long time, and it must simplify creating of  integrations and extensions for the gem. A have PoC, but let's first start with the above. Thanks!

toupeira:
@nbulaj thanks for reaching out! That definitely sounds nice, but to be honest I barely have the time to maintain this gem as it is :crying_cat_face: 

I opened an issue now at https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/89 to find new maintainers.

linhdangduy:
@nbulaj @toupeira 
> a plugin system for Doorkeeper for a long time, and it must simplify creating of integrations and extensions for the gem.

Great to hear that!!! I'm looking forward to your PoC

but before it became existence, could I give a try to update the current doorkeeper-openid-connect to be compatibility with doorkeeper > 5.2 ?

toupeira:
@nbudin if you could give the PR at https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/88 a spin in your app, that would be very much appreciated!

nbudin:
@toupeira It works for me!

toupeira:
@nbudin @Ex-Ark thanks for testing! :pray: 

The changes for Doorkeeper 5.2 are now merged and released as version 1.7.0 of this gem.

PR #85, opened 2019-09-24
Changes for Doorkeeper 5.2

toupeira:
Closes https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/88

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,17 +18,10 @@ script:
   - bundle exec rake spec
 
 env:
-  - rails=4.2.0
   - rails=5.0.0
   - rails=5.2.0
 
 rvm:
-  - 2.3
   - 2.4
   - 2.5
   - 2.6
-
-matrix:
-  exclude:
-    - env: rails=4.2.0
-      rvm: 2.6
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 No changes yet.
 
+## v1.7.0
+
+### Changes
+
+- [#85] This gem now requires Doorkeeper 5.2, Rails 5, and Ruby 2.4
+
 ## v1.6.3 (2019-09-24)
 
 ### Changes
--- a/Gemfile
+++ b/Gemfile
@@ -3,9 +3,6 @@ source 'https://rubygems.org'
 # use Rails version specified by environment
 ENV['rails'] ||= '5.2.0'
 gem 'rails', "~> #{ENV['rails']}"
-
-if ENV['rails'] =~ /^5./
-  gem 'rails-controller-testing'
-end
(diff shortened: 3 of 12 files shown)


linhdangduy:
@toupeira 
I'm the person make some changes at Doorkeeper that mentioned in above issues.
especially on this pull requst: https://github.com/doorkeeper-gem/doorkeeper/pull/1277
Seem like there are other changes, too (like Rails 6!?) but I'm happy to work on this problem.
How could I join?

toupeira:
@linhdangduy thanks for offering your help, that's greatly appreciated! :heart: 

I've extracted Rails 6 support into [its own issue](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/91) and removed it again from `.travis.yml` in this PR. If you want to work on that, you're very welcome to fork and submit a PR.

IIRC the Doorkeeper 5.2 compatibility is pretty much done now in this PR, Travis is passing so I think I'll take another look at it this week and will cut a release.

In the meantime, if you or anybody else could do a review, or test it in your application, that would be very awesome!

Ex-Ark:
I had the same problem as the original issue 
```
wrong number of arguments (given 3, expected 1..2)
doorkeeper (5.2.1) lib/doorkeeper/oauth/pre_authorization.rb:20:in initialize'
doorkeeper-openid_connect (1.6.2) lib/doorkeeper/openid_connect/oauth/pre_authorization.rb:8:in initialize'
```
Using this branch makes the error disappear and the rest of my flow is now working.
Thanks for the fix !

env:
```
rails 6.0
ruby 2.5.7 (mingw32)
doorkeeper-openid_connect (1.7.0 3f53e0f)
```

camilo0365:
Thanks a lot @toupeira !! I will be experimenting with 1.7.0 and get back with feedback.

ansonhoyt:
Thanks for landing this @toupeira! I see the autoload constants deprecation warning is gone, nice! Got my tests passing again and everything is still working for me.

The Doorkeeper migration guide could use more detail, but I was able to work through figuring out the translations changed, migrations changed and scopes are now required (not just for the db, but also auth calls).

Thanks again.

nbulaj:
Oh, thanks @ansonhoyt for writing this. There are many changes and I can't always remember to fill migration guides with all the notes. Maybe you can remember what you've done during migration and add some missed details to upgrade guides ? Would be very helpfull!

Also I think I need to improve PR template :+1:

ansonhoyt:
@nbulaj thank you for your maintenance on such a useful gem!

That's a very fair request. I'll be revisiting our OIDC setup with my current project when working on getting the SSO working, so will keep an eye out for things to put into a PR. I'm missing deadlines on core features right now, so it might be a while 😟

PR #90, opened 2019-10-11
Add authorization_url_host option

steti:
This change adds a new configuration option, `authorization_url_host`.  If set, this causes the discovery endpoint to generate the `authorization_url` using this setting as the `host`.  The rationale is explained in my change to the README:

> This is useful when you want authorization requests to use a different URL than other requests. For example, if your Doorkeeper server is behind a firewall with other servers, you might want other servers to use an "internal" URL to communicate with Doorkeeper, but you want to present an "external" URL to end-users for authentication requests.

This change scratches my particular itch for configuration, but I'm not sure if it makes sense to make other generated URLs configurable as well, or make this configurable in a different way.  I'm happy to raise an issue for discussion instead, but I figured I might as well raise a PR since I already wrote the code :).

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,16 @@ The following settings are optional:
   - Note that the OIDC specification mandates HTTPS, so you shouldn't change this
     for production environments unless you have a really good reason!
 
+- `authorization_url_host`
+  - The hostname to use when generating the `authorization_url` in the discovery response.
+  - The default is to use the request host, just like all the other URLs in the discovery response.
+  - This is useful when you want authorization requests to use a different URL than other requests.
+    For example, if your Doorkeeper server is behind a firewall with other servers, you might want
+    other servers to use an "internal" URL to communicate with Doorkeeper, but you want to present
+    an "external" URL to end-users for authentication requests. Note that this setting does not
+    actually change the URL that your Doorkeeper server responds on - that is outside the scope of
+    Doorkeeper.
+
 ### Scopes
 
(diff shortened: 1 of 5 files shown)


toupeira:
@steti thanks for the PR, and please excuse the delay!

This generally looks great, but I have one thought on this:

It seems to me we should use the custom host in all the URLs returned from the discovery endpoint, since they're all supposed to be called from the client. Do you agree, or would this break something in your use-case? If not, we could generalize this into a `discovery_url_host` setting (or maybe even a `discovery_url_options` hash with a default of `{}`, in case somebody needs e.g. a custom port) and pass it to all the URL helpers.

Let me know if that works for you and if you have time to make these changes, otherwise I'd be happy to finish it! :slightly_smiling_face:

nbulaj:
Hi @steti . Are you still interested in this PR ?

davidbasalla:
This looks really useful! 💯 (and agreed to make _all_ URLs configurable). Is there any progress on this PR?

phlegx:
Exactly what I need! Thx @steti. Can you solve this PR conflicts?

phlegx:
@steti @toupeira @nbulaj @davidbasalla please consider my PR #126. This is a re-work of this PR taking in consideration of your ideas and opinions.

toupeira:
Closing in favour of https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/126

PR #98, opened 2019-12-06
Use cascading deletes for deleting oauth_openid_requests upon deleting an access_grant

### LINKED ISSUE #82: Unable to delete authorized RP application
### Steps to reproduce
- Create an application
- User authorizes application
- Go to doorkeeper admin page and delete application

### Expected behavior
Able to delete application

### Actual behavior
Unable to delete application:
> Mysql2::Error - Cannot delete or update a parent row: a foreign key constraint fails (sample_app.oauth_openid_requests, CONSTRAINT fk_rails_77114b3b09 FOREIGN KEY (access_grant_id) REFERENCES oauth_access_grants (id)):

### Reason
There are two problems.
First the problems above about Mysql, on `doorkeeper-openid-connect` table migration, I found that foreign key is added:
https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/302575597532ecaf5a4ae8000b2c387f7d08371a/lib/generators/doorkeeper/openid_connect/templates/migration.rb.erb#L8-L12

It's off-topic but I'm not understand the purpose to use foreign key, as all the validation can be done on Rails. 

There are associations between: 
- `Doorkeeper::Application` 
- `Doorkeeper::OpenidConnect::AccessGrant`  
- `Doorkeeper::OpenidConnect::Request`

Delete `Application` will delete `AccessGrant` and delete `Request`. However due to foreign key, `AccessGrant` can't be delete without delete associated `Request` first.

That's one problems, we can solve that by deleting foreign key in `templates/migration.rb.erb`, for existing application we can use migration:
```ruby
def up
  remove_foreign_key :oauth_openid_requests, :oauth_access_grants
end

def down
  add_foreign_key(
      :oauth_openid_requests,
      :oauth_access_grants,
      column: :access_grant_id
  )
end
```

No error occurs. However, another problem is that associated `openid_request` are not deleted. To get to the reason, first let's see the model association:

Doorkeeper::Application
https://github.com/doorkeeper-gem/doorkeeper/blob/45b6ce568523a5e2f2cdc9c3ed60254d7891f8dc/lib/doorkeeper/orm/active_record/application.rb#L3-L11
```ruby
class Doorkeeper::Application
    has_many :access_grants, dependent: :delete_all, class_name: "Doorkeeper::AccessGrant"
    has_many :access_tokens, dependent: :delete_all, class_name: "Doorkeeper::AccessToken"
end
```

Doorkeeper::OpenidConnect::AccessGrant
https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/302575597532ecaf5a4ae8000b2c387f7d08371a/lib/doorkeeper/openid_connect/orm/active_record/access_grant.rb#L1-L10

To recap:
```
Application has_many :access_grants, dependent: :delete_all
AccessGrant has_one :openid_request, dependent: :delete
```

When application is deleted, access_grants also are deleted, then openid_request should be deleted as well. However, because we are using `dependent: :delete_all` in `Doorkeeper::Application` so callback stop after delete `:access_grants`. If I change to `dependent: :destroy`, both `:access_grants` **and** `:openid_request` are deleted.

manojmj92:
This fix closes #82

This change works only for new installations for this gem, since the fix is made in the migration template, which is only used during a fresh install.

For existing users who wish to avail the changes to fix the issue, please consider generating a new migration in your application which drops the existing `foreign_key` in the `oauth_openid_requests` table and then creating a new `foreign_key` in it's place with `on_delete: :cascade` included.

Example:

```ruby
class UpdateOauthOpenIdRequestsForeignKeys < ActiveRecord::Migration[5.2]
  def up
    remove_foreign_key(:oauth_openid_requests, column: :access_grant_id)
    add_foreign_key(:oauth_openid_requests, :oauth_access_grants, column: :access_grant_id, on_delete: :cascade)
  end

  def down
    remove_foreign_key(:oauth_openid_requests, column: :access_grant_id)
    add_foreign_key(:oauth_openid_requests, :oauth_access_grants, column: :access_grant_id)
  end
end

```

### DIFF
--- a/lib/generators/doorkeeper/openid_connect/templates/migration.rb.erb
+++ b/lib/generators/doorkeeper/openid_connect/templates/migration.rb.erb
@@ -8,7 +8,8 @@ class CreateDoorkeeperOpenidConnectTables < ActiveRecord::Migration<%= migration
     add_foreign_key(
       :oauth_openid_requests,
       :oauth_access_grants,
-      column: :access_grant_id
+      column: :access_grant_id,
+      on_delete: :cascade
     )
   end
 end
--- a/spec/models/access_grant_spec.rb
+++ b/spec/models/access_grant_spec.rb
@@ -12,4 +12,15 @@ describe Doorkeeper::OpenidConnect::AccessGrant do
       dependent: :delete,
     })
   end
+
+  describe '#delete' do
+    it 'cascades to oauth_openid_requests' do
+      pending('Rails 6 changes - https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/91')
+
+      access_grant = create(:access_grant, application: create(:application))
+      create(:openid_request, access_grant: access_grant)
+
+      expect { access_grant.delete }.to change(Doorkeeper::OpenidConnect::Request, :count).by(-1)
+    end
+  end
 end


toupeira:
@manojmj92 thanks for this PR, and please excuse the long delay :see_no_evil: 

I'll merge this and will check if I can get the pending spec running with Rails 5.

@truongnmt yes great point, I'll add it to the CHANGELOG :+1:

PR #103, opened 2020-02-07
Add support for end_session_endpoint

ralph:
This is in the spec for discovery:
https://openid.net/specs/openid-connect-discovery-1_0.html

And in the spec for RP-Initiated Logout:
https://openid.net/specs/openid-connect-session-1_0.html#RPLogout

And used by popular libraries like:
https://github.com/IdentityModel/oidc-client-js

When you trigger the `userUnloaded` event with `oidc-client-js`, it will print the following error in the browser console when used with `doorkeeper-openid_connect`:

Bildschirmfoto 2020-02-07 um 16 01 45

With the end_session_endpoint configured like in this PR, `oidc-client-js` will work as expected when emitting the `userUnloaded` event (running the client side hook and then redirecting to the specified url).

If you want to use named routes inside the initializer, you can do it like so:

```
  end_session_endpoint do
    Rails.application.routes.url_for({ host: ENV.fetch('HOSTNAME') }, :destroy_user_session)
  end
```

### DIFF
--- a/app/controllers/doorkeeper/openid_connect/discovery_controller.rb
+++ b/app/controllers/doorkeeper/openid_connect/discovery_controller.rb
@@ -30,6 +30,7 @@ module Doorkeeper
           introspection_endpoint: oauth_introspect_url(protocol: protocol),
           userinfo_endpoint: oauth_userinfo_url(protocol: protocol),
           jwks_uri: oauth_discovery_keys_url(protocol: protocol),
+          end_session_endpoint: Doorkeeper::OpenidConnect.configuration.end_session_endpoint.call(),
 
           scopes_supported: doorkeeper.scopes,
 
@@ -67,7 +68,7 @@ module Doorkeeper
             'exp',
             'iat',
           ] | openid_connect.claims.to_h.keys,
-        }
+        }.compact
       end
 
       def webfinger_response
--- a/README.md
+++ b/README.md
@@ -150,6 +150,10 @@ The following settings are optional:
   - Note that the OIDC specification mandates HTTPS, so you shouldn't change this
     for production environments unless you have a really good reason!
 
+- `end_session_endpoint`
+  - The URL that the user is redirected to after ending the session on the client.
+  - Used by implementations like https://github.com/IdentityModel/oidc-client-js.
+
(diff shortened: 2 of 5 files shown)


app/controllers/doorkeeper/openid_connect/discovery_controller.rb
```diff
@@ -30,6 +30,7 @@ def provider_response
           introspection_endpoint: oauth_introspect_url(protocol: protocol),
           userinfo_endpoint: oauth_userinfo_url(protocol: protocol),
           jwks_uri: oauth_discovery_keys_url(protocol: protocol),
+          end_session_endpoint: Doorkeeper::OpenidConnect.configuration.end_session_endpoint.call(),
```
giulliano-bueno: Is this part exercised by a test? Could you point me the file, please? Thanks.

ralph: I added a spec that this is added to the configuration here:
https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/103/files#diff-db9ee42799b71b9a732daef09d2ebdcf

And another one for the controller to make sure it returns the value in the JSON response:
https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/103/files#diff-d828e386ea3f9924e9fda0d952d9e561

Also, after you asking, I also added another spec to make sure that no end_session_endpoint is present in the discovery response if none was configured.

Am I missing anything?

nbulaj, line 33: Could we please remove `()`after the `#call` ?

nbulaj: Also you could reuse `openid_connect` instead of `::Doorkeeper::OpenidConnect.configuration` (see above lines)

nbulaj: Will do myself

ralph: Thank you so much! ✌️ 

Thank you for contributing so much to open source! 👍

PR #111, opened 2020-05-23
Add configuration for prompt=select_account

### LINKED ISSUE #107: Configuration method for prompt=select_account
### Problem
With `prompt=login`, we have the [reauthenticate_oidc_resource_owner](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/v1.7.1/lib/doorkeeper/openid_connect/helpers/controller.rb#L64-L65) method to call the config [reauthenticate_resource_owner](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/v1.7.1/lib/doorkeeper/openid_connect/config.rb#L112-L114). So user just have to defined some code in `reauthenticate_resource_owner` in doorkeeper_openid_connect.rb to execute the code as user's need.

Why doorkeeper oidc don't have the equivalent config when `prompt=select_account`?
https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/lib/doorkeeper/openid_connect/helpers/controller.rb#L68-L71
If possible could we add config function for `select_account`?

### Currently behavior
Because the gem only write that:
```ruby
when 'select_account'
  # TODO: let the user implement this
  raise Errors::AccountSelectionRequired
```
the error is rescued at [here](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/v1.7.1/lib/doorkeeper/openid_connect/helpers/controller.rb#L14-L16). So to be able to use `prompt=select_account`, I see that user have to monkey-patch the [handle_oidc_prompt_param!](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/v1.7.1/lib/doorkeeper/openid_connect/helpers/controller.rb#L55) method as following:
```ruby
def handle_oidc_prompt_param!(owner)
  super(owner)
rescue Errors::AccountSelectionRequired
  # user's defined select_account behavior  
end
```
that not clean, hard to maintain for user.

### Expect
```ruby
when 'select_account'
  call_select_account_config(owner) # execute the config for select_account
```

What do you think about this idea?

linhdangduy:
## Summary
Add `select_account_for_resource_owner` configuration to allow defining behavior when receiving param `prompt=select_account` at authorization request.

Resolves #107

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - [#114] Fix user_info endpoint when used in api mode
 - [#112] Add grant_types_supported to discovery response
+- [#111] Add configuration callbacks `select_account_for_resource_owner` to support param prompt=select_account
 
 ## v1.7.2 (2020-05-20)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 - [#114] Fix user_info endpoint when used in api mode
 - [#112] Add grant_types_supported to discovery response
-- [#111] Add configuration callbacks `select_account_for_resource_owner` to support param prompt=select_account
+- [#111] Add configuration callback `select_account_for_resource_owner` to support the `prompt=select_account` param
 
 ## v1.7.2 (2020-05-20)
 
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ The following settings are optional, but recommended for better client compatibi
   - Required to support the `max_age` and `prompt=login` parameters.
   - The block is executed in the controller's scope, so you have access to methods like `params`, `redirect_to` etc.
 - `select_account_for_resource_owner`
-  - Defines how to trigger account selecting to choose the current login user
+  - Defines how to trigger account selection to choose the current login user.
   - Required to support the `prompt=select_account` parameter.
   - The block is executed in the controller's scope, so you have access to methods like `params`, `redirect_to` etc.
 
--- a/README.md
+++ b/README.md
@@ -139,6 +139,10 @@ The following settings are optional, but recommended for better client compatibi
   - Defines how to trigger reauthentication for the current user (e.g. display a password prompt, or sign-out the user and redirect to the login form).
   - Required to support the `max_age` and `prompt=login` parameters.
   - The block is executed in the controller's scope, so you have access to methods like `params`, `redirect_to` etc.
+- `select_account_for_resource_owner`
+  - Defines how to trigger account selecting to choose the current login user
+  - Required to support the `prompt=select_account` parameter.
+  - The block is executed in the controller's scope, so you have access to methods like `params`, `redirect_to` etc.
 
 The following settings are optional:
 
--- a/lib/doorkeeper/openid_connect/helpers/controller.rb
+++ b/lib/doorkeeper/openid_connect/helpers/controller.rb
@@ -96,7 +96,7 @@ module Doorkeeper
           end
         end
 
-        def return_without_prompt_param(prompt_value)
+        def return_without_oidc_prompt_param(prompt_value)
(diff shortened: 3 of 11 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - [#114] Fix user_info endpoint when used in api mode
+- [#111] Add configuration callbacks `select_account_for_resource_owner` to support param prompt=select_account
```
toupeira, line 4:
```suggestion
- [#111] Add configuration callback `select_account_for_resource_owner` to support the `prompt=select_account` param
```

README.md
```diff
@@ -139,6 +139,10 @@ The following settings are optional, but recommended for better client compatibi
   - Defines how to trigger reauthentication for the current user (e.g. display a password prompt, or sign-out the user and redirect to the login form).
   - Required to support the `max_age` and `prompt=login` parameters.
   - The block is executed in the controller's scope, so you have access to methods like `params`, `redirect_to` etc.
+- `select_account_for_resource_owner`
+  - Defines how to trigger account selecting to choose the current login user
```
toupeira, line 143:
```suggestion
  - Defines how to trigger account selection to choose the current login user.
```

lib/doorkeeper/openid_connect/helpers/controller.rb
```diff
@@ -75,8 +75,7 @@ def handle_oidc_prompt_param!(owner)
             when 'consent'
               render :new
             when 'select_account'
-              # TODO: let the user implement this
-              raise Errors::AccountSelectionRequired
```
toupeira, line 79: Can you please remove the now unused `Errors::AccountSelectionRequired` in `lib/doorkeeper/openid_connect/errors.rb`?

lib/doorkeeper/openid_connect/helpers/controller.rb
```diff
@@ -97,16 +96,21 @@ def handle_oidc_max_age_param!(owner)
           end
         end
 
-        def reauthenticate_oidc_resource_owner(owner)
+        def return_without_prompt_param(prompt_value)
```
toupeira, line 99: Worried about name clashes again:

```suggestion
        def return_without_oidc_prompt_param(prompt_value)
```

(maybe at some point we can move all these helper methods into a module that doesn't get included, so this stops being a concern :wink:)

lib/generators/doorkeeper/openid_connect/templates/initializer.rb
```diff
@@ -28,6 +28,18 @@
     # redirect_to new_user_session_url
   end
 
+  # Depend on the configuration, the DoubleRenderError could be occurred
+  #  because of render/redirect is called at some configuration before this configure is called.
+  # To avoid DoubleRenderError, you could adding these two lines at the beginning
+  #  of this configuration: (Reference: https://github.com/rails/rails/issues/25106)
```
toupeira, line 34:
```suggestion
  # Depending on your configuration, a DoubleRenderError could be raised
  # if render/redirect_to is called at some point before this callback is executed.
  # To avoid the DoubleRenderError, you could add these two lines at the beginning
  #  of this callback: (Reference: https://github.com/rails/rails/issues/25106)
```

lib/doorkeeper/openid_connect/helpers/controller.rb
linhdangduy: https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/111/commits/ddc0ede949f40e745ed51c49f96826da6adb165f#diff-b2596ff7dd36df28fc1ab37344ff56a9L29
Removed it 🙆‍♂️

toupeira:
@linhdangduy thanks for your changes, this is looking great now! :tada: 

I'll go ahead and merge...

PR #113, opened 2020-05-24
API mode will not define the verify_authenticity_token action

### LINKED ISSUE #114: userinfo endpoint broken in rails api mode
Rails API base controller does not include the "verify_authenticity_token" method, which means the skip check in the userinfo controller raises an exception

jmortlock:
### WHY

If you run this in a Rails 5 API application you will get an <ArgumentError: Before process_action callback :verify_authenticity_token has not been defined> error

Resolves #114

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- [#114] Fix user_info endpoint when used in api mode
+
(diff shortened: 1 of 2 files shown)


nbulaj:
Could you please add a changelog entry?

jmortlock:
Changelog added

PR #116, opened 2020-05-29
Register custom Grant Flows using new Doorkeeper API

nbulaj:
Doorkeeper 5.5 would have a new public API to register custom grant flows - see https://github.com/doorkeeper-gem/doorkeeper/pull/1418.

This is preventive PR to have support for it and not to see deprecation messages.

But I have some doubts about the changes so would be great to have any feedback as I'm not familiar with OpenID connect :smiley: :pray:


spec/controllers/discovery_controller_spec.rb
```diff
@@ -19,7 +19,7 @@
 
         'scopes_supported' => ['openid'],
 
-        'response_types_supported' => ['code'],
+        'response_types_supported' => ['code', 'token', 'id_token', 'id_token token'],
```
nbulaj, line 22: Is it OK that before response types supported didn't have this values? Or it's my stupid mistake and I need to register this custom grant flows not as authorization, but as token grants?

lib/doorkeeper/openid_connect.rb
```diff
@@ -63,5 +63,33 @@ def self.signing_key_normalized
         key.slice(:kty, :kid)
       end
     end
+
+    if defined?(::Doorkeeper::GrantFlow)
+      Doorkeeper::GrantFlow.register(
+        :token,
```
nbulaj, line 69: Also I need to register it, but it seems to me that when we add `token` to `grant_flows` (here https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/lib/doorkeeper/openid_connect/response_types_config.rb#L9) we meant `implicit` grant, right? Or correct me please

toupeira: I'm not sure, this was implemented in https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/45 and I think the idea was that you might want to enable Implicit Flow for normal OAuth uses, but not for OIDC (or vice versa). Maybe @nashby can shine some light on this? :grinning:

I'm also not sure if `response_type=token` by itself is valid in an OIDC context, the spec has this table at https://openid.net/specs/openid-connect-core-1_0.html#Authentication:

"response_type" value | Flow
-- | --
code | Authorization Code Flow
id_token | Implicit Flow
id_token token | Implicit Flow
code id_token | Hybrid Flow
code token | Hybrid Flow
code id_token token | Hybrid Flow

But maybe we're adding it in https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/lib/doorkeeper/openid_connect/response_types_config.rb#L6-L18 in case only `implicit_oidc` is enabled but not Doorkeeper's `implicit` flow, to allow the `code token` combination? If we unified `implicit` and `implicit_oidc` this probably wouldn't be necessary.

In any case, it looks like we can remove the definition for `Doorkeeper::Request::Token` here since we're already adding it in https://github.com/doorkeeper-gem/doorkeeper/blob/b71c5afbfb6694cc023e30be84b9018b5280bf0e/lib/doorkeeper/grant_flow.rb#L11-L15, though with the name `:implicit` instead of `:token`?

spec/controllers/discovery_controller_spec.rb
toupeira: @nbulaj good catch! :thinking: 

I think they should only be included if implicit flows are enabled, which doesn't seem to be the case in either the default settings or our spec dummy app. (and we should add specs for this case :grin:)

So previously we were including them conditionally in https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/116/files#diff-0c58b0448c601da9348f2f321985fe19, but now with the new `Doorkeeper::GrantFlow` they're always included?

This definitely smells like a bug somewhere. Maybe the problem is that we're unconditionally adding the aliased grants at https://github.com/doorkeeper-gem/doorkeeper/blob/b71c5afbfb6694cc023e30be84b9018b5280bf0e/lib/doorkeeper/config.rb#L600?

nbulaj: Whithout aliases Doorkeeper would include all the OID grants in the list of enabled... Yep, there definitely could be some issue and I need to find it before the release :hand_over_mouth:

lib/doorkeeper/openid_connect.rb
nbulaj: Yep, I just wanted to be sure it's similar grants :)

spec/controllers/discovery_controller_spec.rb
nbulaj: > So previously we were including them conditionally in https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/116/files#diff-0c58b0448c601da9348f2f321985fe19, but now with the new Doorkeeper::GrantFlow they're always included?

Previously it was like this:

https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/lib/doorkeeper/openid_connect/response_types_config.rb#L6-L18

So if user configured Doorkeeper as:

```ruby
Doorkeeper.configure do

  grant_flows ["implicit_oidc"]
end
```

It converted to three more grant flows and [resolved by `Doorkeeper::Request` class](https://github.com/doorkeeper-gem/doorkeeper/blob/8fec40d5b873dc62e8bc29b8ebfe71a003b09dab/lib/doorkeeper/request.rb#L30-L33).

But now with `GrantFlow` API all the custom grant flows (like `id_token`) must be explicitly configured and registered, so Doorkeeper could understand:

* what is `implicit_oidc` grant flow
* and how to process it

Since `implicit_oidc` is actually an alias (or how to name it?) to three other grants - we must consider this case.

nbulaj: Oh, actually this is the issue!

> Maybe the problem is that we're unconditionally adding the aliased grants at

I'll fix it, thannx for fresh eyes :D :+1:

nbulaj: OK, I fixed Doorkeeper itself and defined grant flows in dummy app for specs to test

PR #118, opened 2020-06-04
Bug fix: Fragment urls for implicit flow error redirects

joeljunstrom:
We noticed that when using the `promt=none` option our client did not parse the returned login error.

This was due to that the computed redirect url for errors used query strings instead of fragments even if the response_type was of the id_token type. 

This PR aims to ensure that the same response mode is used regardless on success or error as per https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthError 

> If the End-User denies the request or the End-User authentication fails, the Authorization Server MUST return the error Authorization Response in the fragment component of the Redirection URI …  unless a different Response Mode was specified.

and https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Combinations

> When supplied as the value for the response_type parameter, a successful response MUST include an Access Token, an Access Token Type, and an id_token. The default Response Mode for this Response Type is the fragment encoding and the query encoding MUST NOT be used. Both successful and error responses SHOULD be returned using the supplied Response Mode, or if none is supplied, using the default Response Mode.

It _does not_ try to fix responding to the response mode parameter as this seems to be unsupported by doorkeeper at the moment (unless I'm misstaken, it has been known to happen before).

Further refactoring could be made to ensure the error responses is not as tied to the pre_auth as it is currently but I tried to keep the change small.

Do note that this is the first time I've touched openid / oauth stuff so sorry if I misunderstood how this is supposed to work! =)

### DIFF
--- a/lib/doorkeeper/openid_connect/helpers/controller.rb
+++ b/lib/doorkeeper/openid_connect/helpers/controller.rb
@@ -49,7 +49,7 @@ module Doorkeeper
                                name: exception.type,
                                state: params[:state],
                                redirect_uri: params[:redirect_uri],
-                               response_on_fragment: pre_auth.response_on_fragment?,
+                               response_on_fragment: Doorkeeper::OpenidConnect::ResponseMode.new(pre_auth.response_type).fragment?,
                              )
           end
 
--- a/lib/doorkeeper/openid_connect/helpers/controller.rb
+++ b/lib/doorkeeper/openid_connect/helpers/controller.rb
@@ -38,18 +38,20 @@ module Doorkeeper
           # FIXME: workaround for Rails 5, see https://github.com/rails/rails/issues/25106
           @_response_body = nil
 
+          should_respond_on_fragment = Doorkeeper::OpenidConnect::ResponseMode.new(pre_auth.response_type).fragment?
           error_response = if exception.type == :invalid_request
                              ::Doorkeeper::OAuth::InvalidRequestResponse.new(
                                name: exception.type,
                                state: params[:state],
                                redirect_uri: params[:redirect_uri],
+                               response_on_fragment: should_respond_on_fragment,
                              )
                            else
                              ::Doorkeeper::OAuth::ErrorResponse.new(
                                name: exception.type,
                                state: params[:state],
                                redirect_uri: params[:redirect_uri],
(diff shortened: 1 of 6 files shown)


lib/doorkeeper/openid_connect/helpers/controller.rb
```diff
@@ -38,17 +38,20 @@ def handle_oidc_error!(exception)
           # FIXME: workaround for Rails 5, see https://github.com/rails/rails/issues/25106
           @_response_body = nil
 
+          should_respond_on_fragment = Doorkeeper::OpenidConnect::ResponseMode.new(pre_auth.response_type).fragment?
```
toupeira, line 41: Could we just call `pre_auth.response_on_fragment?` here, and make that method public?

lib/doorkeeper/openid_connect/response_mode.rb
```diff
@@ -0,0 +1,30 @@
...
+        mode == 'fragment'
+      end
+
+      def query?
+        mode == 'query'
+      end
+
+      def mode
+        case type
+        when 'token', 'id_token', 'id_token token'
```
toupeira, line 22: Trying to double-check the specs here:

- `id_token` and `id_token token` are used for OIDC Implicit Flow which mandates `fragment`: https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthRequest
- `token` is used in normal OAuth and https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#ResponseModes specifies `fragment` as the default.

So I think this looks correct! :+1:

lib/doorkeeper/openid_connect/helpers/controller.rb
joeljunstrom: Sure! I didn't want to mess about with the api given my low familiarity with the code base. I'll fix that.

lib/doorkeeper/openid_connect/response_mode.rb
joeljunstrom: Yea I think it should be according to spec, nice that you double checked. Note that I think that I should try and upstream the response_mode as part of the new ”Grant Flows” in doorkeeper (since it makes more sense that they dictate what is default and/or valid response modes. 

But I think this is an ok change until it is available from there.

PR #119, opened 2020-06-08
Execute end_session_endpoint in the controllers context

joeljunstrom:
Allows the application to use the configured host etc in url builders without having to set them again for the specific call.

Rationale is that it is cumbersome enough as it is to maintain this for mailers

### DIFF
--- a/app/controllers/doorkeeper/openid_connect/discovery_controller.rb
+++ b/app/controllers/doorkeeper/openid_connect/discovery_controller.rb
@@ -32,7 +32,7 @@ module Doorkeeper
           introspection_endpoint: oauth_introspect_url(protocol: protocol),
           userinfo_endpoint: oauth_userinfo_url(protocol: protocol),
           jwks_uri: oauth_discovery_keys_url(protocol: protocol),
-          end_session_endpoint: openid_connect.end_session_endpoint.call,
+          end_session_endpoint: openid_connect.end_session_endpoint.call(self),
 
           scopes_supported: doorkeeper.scopes,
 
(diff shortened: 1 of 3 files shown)


app/controllers/doorkeeper/openid_connect/discovery_controller.rb
```diff
@@ -32,7 +32,7 @@ def provider_response
           introspection_endpoint: oauth_introspect_url(protocol: protocol),
           userinfo_endpoint: oauth_userinfo_url(protocol: protocol),
           jwks_uri: oauth_discovery_keys_url(protocol: protocol),
-          end_session_endpoint: openid_connect.end_session_endpoint.call,
+          end_session_endpoint: openid_connect.end_session_endpoint.call(self),
```
toupeira, line 35: I think this will crash if somebody defines a block in their initializer without arguments.

We could instead just use `instance_exec` and run the block in the scope of the controller:

```suggestion
          end_session_endpoint: instance_exec(&openid_connect.end_session_endpoint),
```

And then in the spec below we can just call `logout_url` directly.

joeljunstrom: Well spotted, applied your suggestion now

joeljunstrom:
@toupeira did we miss to actually merge this before the release commit? =)

toupeira:
@joeljunstrom oh yikes, I totally did. :facepalm: Sorry about that!

Merged and released in 1.7.4 now! :grin:

joeljunstrom:
Haha no worries <3

PR #123, opened 2020-07-18
Remove reference to ApplicationRecord

wheeyls:
This reference to `ApplicationRecord` is causing an error when initializing in Rails 6:

    DEPRECATION WARNING: Initialization autoloaded the constant ApplicationRecord. (ActiveSupport::DeprecationException)

This is a change in the way Rails 6 does autoloading; it looks like this was done to appease rubocop's style rules, which shouldn't apply to this situation.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -54,6 +54,8 @@ Rails/RakeEnvironment:
     - Rakefile
 Rails/SkipsModelValidations:
   Enabled: false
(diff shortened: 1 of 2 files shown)


toupeira:
@wheeyls thanks! :+1: 

This LGTM so I'll go ahead and merge...

/cc @nbulaj

PR #138, opened 2021-01-01
Bump doorkeeper 5.5 and define body method for authorization response to support response_mode=form_post

### LINKED ISSUE #92: Is the form_post response_mode supported?
It is my understanding that if the response type is `form_post` the OP will POST the response instead of encoding the id_token (etc) in the query string or fragment.

I can't find evidence of it being supported by looking at the source code.

Please advise, TIA.

linhdangduy:
## Description
Resolves #92 

Because [doorkeeper already supported response_mode parameter on master branch](https://github.com/doorkeeper-gem/doorkeeper/pull/1438), so I create this pull to support this response_mode=form_post on doorkeeper-openid_connect. 

## What did I do

- OAuth::IdTokenResponse and OAuth::IdTokenTokenResponse:
  - Define `body` method [to return the result when in response_mode=form_post](https://github.com/doorkeeper-gem/doorkeeper/blob/b044b3333755fb87423091d0a5faff7d770764d1/app/controllers/doorkeeper/authorizations_controller.rb#L57)
  - add response_mode_matches to `id_token` and `id_token token` response_type.
- add CHANGELOG.md
- add [OAuth 2.0 Form Post Response Mode](https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html) to README

## ~Caution ⚠️~

~~Doorkeeper will released the response_mode=form_post from v5.5.0, these PR should be merged and released at that time.~~

**Update**

Doorkeeper released v5.5.0! So this pull is ready to be merged 🎉


doorkeeper-openid_connect.gemspec
```diff
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.4'
 
-  spec.add_runtime_dependency 'doorkeeper', '>= 5.2', '< 5.5'
+  spec.add_runtime_dependency 'doorkeeper', '>= 5.5', '< 5.6'
```
nbulaj, line 27: Do we need such strict dependency ? This extension currently supports previous versions of Doorkeeper (check  lib/doorkeeper/openid_connect.rb#63 - there is a check `if defined?(::Doorkeeper::GrantFlow)`)

nbulaj: @linhdangduy could you please address this? :)

linhdangduy: > Do we need such strict dependency ? This extension currently supports previous versions of Doorkeeper (check lib/doorkeeper/openid_connect.rb#63 - there is a check if defined?(::Doorkeeper::GrantFlow))

Sorry for the delay.

Yes, we need.

The change that I made was adding [response_mode_matches](https://github.com/doorkeeper-gem/doorkeeper/pull/1438/files#diff-fbf6e8847bf493e8a4c566038f5bec7700e78737b1af85723ddf5790a5856375R16) attribute to `::Doorkeeper::GrantFlow`.  That change is released in v5.5.
So the doorkeeper version below v5.5, with `::Doorkeeper::GrantFlow` defined, does not have that attribute, causing the error we met.

nbulaj: OK, let's call @toupeira  also for a review :)

toupeira: @nbulaj @linhdangduy thanks for the ping!

I'm fine with requiring Doorkeeper 5.5, as the two gems are very tightly coupled anyway :+1:

I see we can remove the code at https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/805b7ca9c42a011e357b2d6daac88c45e7f2172a/lib/doorkeeper/openid_connect.rb#L85-L87 and the file https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/805b7ca9c42a011e357b2d6daac88c45e7f2172a/lib/doorkeeper/openid_connect/response_types_config.rb, @linhdangduy do you want to do that in this PR while we're at it? Otherwise I'd also be happy to do it in a follow-up.

lib/doorkeeper/openid_connect.rb
```diff
@@ -69,12 +69,14 @@ def self.signing_key_normalized
...
         :id_token,
         response_type_matches: 'id_token',
+        response_mode_matches: %w[fragment form_post],
         response_type_strategy: Doorkeeper::OpenidConnect::IdToken,
       )
 
       Doorkeeper::GrantFlow.register(
         'id_token token',
         response_type_matches: 'id_token token',
+        response_mode_matches: %w[fragment form_post],
```
toupeira, line 79: **question:** Do we need to add `query` as well? It's currently listed at https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/805b7ca9c42a011e357b2d6daac88c45e7f2172a/app/controllers/doorkeeper/openid_connect/discovery_controller.rb#L41

We should also add `form_post` in the discovery response there, maybe we can do it dynamically with something like this:

```diff
diff --git i/app/controllers/doorkeeper/openid_connect/discovery_controller.rb w/app/controllers/doorkeeper/openid_connect/discovery_controller.rb
index c6fa4c3..09e139f 100644
--- i/app/controllers/doorkeeper/openid_connect/discovery_controller.rb
+++ w/app/controllers/doorkeeper/openid_connect/discovery_controller.rb
@@ -38,7 +38,7 @@ module Doorkeeper
 
           # TODO: support id_token response type
           response_types_supported: doorkeeper.authorization_response_types,
-          response_modes_supported: %w[query fragment],
+          response_modes_supported: response_modes_supported(doorkeeper),
           grant_types_supported: grant_types_supported(doorkeeper),
 
           # TODO: look into doorkeeper-jwt_assertion for these
@@ -70,6 +70,10 @@ module Doorkeeper
         }.compact
       end
 
+      def response_modes_supported(doorkeeper)
+        doorkeeper.authorization_response_flows.flat_map(&:response_mode_matches).uniq
+      end
+
       def grant_types_supported(doorkeeper)
         grant_types_supported = doorkeeper.grant_flows.dup
         grant_types_supported << 'refresh_token' if doorkeeper.refresh_token_enabled?
```

README.md
```diff
@@ -34,6 +34,7 @@ The following parts of [OpenID Connect Core 1.0](http://openid.net/specs/openid-
 - [Requesting Claims using Scope Values](http://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims)
 - [UserInfo Endpoint](http://openid.net/specs/openid-connect-core-1_0.html#UserInfo)
 - [Normal Claims](http://openid.net/specs/openid-connect-core-1_0.html#NormalClaims)
+- (From doorkeeper v5.5.0) [OAuth 2.0 Form Post Response Mode](https://openid.net/specs/oauth-v2-form-post-response-mode-1_0.html)
```
toupeira, line 37: **nitpick:** Not sure if we should specify the Doorkeeper version here in the README, as we don't do it for other features either :wink: 

But we should mention it in the CHANGELOG, something like `This gem now requires Doorkeeper 5.5`.

toupeira (changes requested):
Overall this LGTM, thanks for the contribution! :heart: 

I've left some minor comments, please take another look.

doorkeeper-openid_connect.gemspec
linhdangduy: @toupeira Thanks for the review. I will take a look today.

lib/doorkeeper/openid_connect.rb
linhdangduy: @toupeira 

> question: Do we need to add query as well? It's currently listed at

The `id_token token` response_type indicates that the implicit grant flow is used. So we _must not_ add `query` response mode to implicit grant flow.

That is written in openid spec as following. https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Combinations
> id_token token
...The **default Response Mode** for this Response Type **is the fragment encoding** and **the query encoding MUST NOT be used**...

(I also mentioned it in [this pull](https://github.com/doorkeeper-gem/doorkeeper/pull/1438) )
> Implicit Flow:
"query" is not allow because with this response_mode, sensitive data (access token, id token (oidc)) can be seen by user/browser history/server log)

doorkeeper-openid_connect.gemspec
linhdangduy: https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/138/commits/99ef25ed9b6949c64d5aff08e7b29e6349cb393e

I removed that unnecessary part.

README.md
linhdangduy: It fixed! https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/138/commits/99ef25ed9b6949c64d5aff08e7b29e6349cb393e

lib/doorkeeper/openid_connect.rb
linhdangduy: I added response_modes_supported and test at here https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/138/commits/5b5b4695229b1d878c50b00a1b510fcae69ce262

nbulaj (approved):
Good work :handshake:

linhdangduy:
@nbulaj @toupeira 

After fixing the comments of @toupeira, I added more two commits. Explaining below:

1. https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/138/commits/0b681cb389c1945d73362895c0fed9d17871b4eb `delete unnecessary support for dookeeper < v5.5` 
Because the minimize doorkeeper's version is v5.5 now, so we can drop the code support for Doorkeeper version that less than v5.5
2. https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/138/commits/d07ba31a80699e67db7176b20269c18dc3055556 `update: using response_mode which supported by doorkeeper v5.5`
- Currently, this gem has a class (`doorkeeper/openid_connect/response_mode`) responsible for getting response_mode from response_type. It is added in [this pull](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/118). But it is obsolete because `response_mode` already be supported from doorkeeper v5.5
 - So I deleted the above class, and fix `response_on_fragment?` method in `lib/doorkeeper/openid_connect/oauth/pre_authorization.rb` file, which getting response_mode by using `authorization_response_flows` in doorkeeper config.  (With a comment show that this implementation [need to be used in doorkeeper later](https://github.com/doorkeeper-gem/doorkeeper/blob/v5.5.0/lib/doorkeeper/oauth/pre_authorization.rb#L152-L155), so we don't need overwrite this method)
- Beside that, in the method for handling oidc errors (`handle_oidc_error!`) in `lib/doorkeeper/openid_connect/helpers/controller.rb` file, instead of defining how to response in this gem, I just use the method `redirect_or_render` which [is already implemented in doorkeeper](https://github.com/doorkeeper-gem/doorkeeper/blob/v5.5.0/app/controllers/doorkeeper/authorizations_controller.rb#L52). A `@authorize_response` instance variable for `authorizations_controller` need to be defined, because when response_mode is form_post, `@authorize_response` [is used in form_post template](https://github.com/doorkeeper-gem/doorkeeper/blob/v5.5.0/app/views/doorkeeper/authorizations/form_post.html.erb#L7).

Please leave questions if there are somethings unclear. Thanks!

lib/doorkeeper/openid_connect.rb
toupeira: > The id_token token response_type indicates that the implicit grant flow is used. So we must not add query response mode to implicit grant flow.

Thanks for checking! :+1:

toupeira (approved):

toupeira:
@linhdangduy sorry for the delay, this fell off my radar again :see_no_evil: 

This all LGTM now, thanks a lot for your work on this, and also the cleanups! :heart: 

I'll merge this and will prepare a new release shortly, I'll bump the version to 1.8.0 due to the Doorkeeper 5.5 requirement. :+1:

PR #144, opened 2021-04-19
Update issuer configuration to allow dynamic values

maxxsnake:
Proposing a small change here so that `issuer` accepts a string or a block. This is useful in cases where Doorkeeper API mode is enabled and requests are proxied from multiple hosts, so that the value may be set dynamically from the resource owner and/or application. Please let me know if you have any concerns. Thanks!


toupeira:
@maxxsnake thanks, this LGTM and makes sense! :+1: 

I'll go ahead and merge this, and will also update the README to mention the block syntax.

PR #147, opened 2021-04-20
Release version 1.8.0-rc1

toupeira:
### Upgrading

This gem now requires Doorkeeper 5.5 and Ruby 2.5.

### Changes

- [#138] Support form_post response mode (thanks to @linhdangduy)
- [#144] Support block syntax for `issuer` configuration (thanks to @maxxsnake)
- [#145] Register token flows with the strategy instead of the token class (thanks to @paukul)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,16 @@
 ## Unreleased
 
-- [#138] Support form_post response mode (This gem now requires Doorkeeper v5.5)
-- [#146] Require at least Ruby 2.5
+## v1.8.0-rc1 (2021-04-20)
+
+### Upgrading
+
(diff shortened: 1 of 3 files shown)


toupeira:
@nbulaj FYI, I'll go with an RC because I haven't actually tested this in a real-world app yet :grin: 

If nothing major crops up I'll release a proper 1.8.0 in a couple weeks :crossed_fingers:

toupeira:
@nhosoya I was waiting to see if any new issues are reported, but I guess silence is a good sign too :grin:

https://rubygems.org/gems/doorkeeper-openid_connect/versions/1.8.0.pre.rc1 shows 1000+ downloads, so there seem to be at least *some* users. I'll go ahead and release 1.8.0 now! :+1:

PR #152, opened 2021-06-04
Exposing oauth-authorization-server, too

mitar:
See [this GitLab issue](https://gitlab.com/gitlab-org/gitlab/-/issues/233956) for more background.

In short, both of those URIs and compatible to each other and contain a similar document with possibly slightly different fields, but both allow other fields to be present. So I propose we expose the same document on both locations.

Current document is already a mix of both specs, for example, `revocation_endpoint` is in fact defined for `oauth-authorization-server` and not `openid-configuration`, but it is exposed under `openid-configuration`. Which is fine.

So, in short, for maximum compatibility, let's expose this on both paths.

### DIFF
--- a/lib/doorkeeper/openid_connect/rails/routes.rb
+++ b/lib/doorkeeper/openid_connect/rails/routes.rb
@@ -62,6 +62,7 @@ module Doorkeeper
         def discovery_well_known_routes
           routes.scope path: '.well-known' do
             routes.get :provider, path: 'openid-configuration'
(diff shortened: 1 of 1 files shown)


mitar:
Any update here?

nbulaj (approved):
:+1: LGTM

Could you please add a note to the changelog? Thanks!

nbulaj:
I'll do it myself, thanks

mitar:
Thanks! Sorry, I was busy and could not get to it immediately. Thank you for picking it up.

Issue #151, opened 2021-06-02
/oauth/token returns an ArgumentError in 1.8.0


MH4GF:
Hi!
After upgrading to 1.8.0, I noticed an ArgumentError in /oauth/token .

```
app_1       | Started POST "/oauth/token" for 172.20.0.1 at 2021-06-02 14:56:47 +0900
app_1       | Processing by Doorkeeper::TokensController#create as */*
app_1       |   Parameters: {"client_id"=>"uid", "client_secret"=>"[FILTERED]", "grant_type"=>"password", "username"=>"foo@example.com", "password"=>"[FILTERED]", "scope"=>"openid email", "token"=>{"client_id"=>"uid", "client_secret"=>"[FILTERED]", "grant_type"=>"password", "username"=>"foo@example.com", "password"=>"[FILTERED]", "scope"=>"openid email"}}
app_1       |   Account Load (0.8ms)  SELECT `accounts`.* FROM `accounts` WHERE `accounts`.`email` IS NULL LIMIT 1
app_1       |   ↳ config/initializers/doorkeeper.rb:22:in `block (2 levels) in <main>'
app_1       | Completed 500 Internal Server Error in 3ms (ActiveRecord: 0.8ms | Allocations: 1388)
app_1       |
app_1       |
app_1       | wrong number of arguments (given 5, expected 3..4) excluded from capture: No host specified, no public_key specified, no project_id specified
app_1       |
app_1       | ArgumentError (wrong number of arguments (given 5, expected 3..4)):
app_1       |
app_1       | doorkeeper-openid_connect (1.8.0) lib/doorkeeper/openid_connect/oauth/password_access_token_request.rb:9:in `initialize'
app_1       | doorkeeper (5.5.1) lib/doorkeeper/request/password.rb:9:in `new'
app_1       | doorkeeper (5.5.1) lib/doorkeeper/request/password.rb:9:in `request'
app_1       | doorkeeper (5.5.1) lib/doorkeeper/request/strategy.rb:8:in `authorize'
app_1       | doorkeeper (5.5.1) app/controllers/doorkeeper/tokens_controller.rb:133:in `authorize_response'
app_1       | doorkeeper (5.5.1) app/controllers/doorkeeper/tokens_controller.rb:8:in `create'
app_1       | actionpack (6.0.3.7) lib/action_controller/metal/basic_implicit_render.rb:6:in `send_action'
[...]

PR #153, opened 2021-06-15
Update password_access_token_request.rb to match method parameters in doorkeeper 5.5

CircumnavigatingFlatEarther:
doorkeeper 5.5 added a new parameter `credentials` to `Doorkeeper::Oauth::PasswordAccessTokenRequest.initialize` resulting in ArgumentError thrown. See [this issue](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/151).

Once we match the parameters in the methods, we can set `skip_client_authentication_for_password_grant` to `true` and it works just fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- [#153] Fix ArgumentError caused by client credential validation introduced in Doorkeeper 5.5.1
+
 ## v1.8.0 (2021-05-11)
 
 No changes from v1.8.0-rc1.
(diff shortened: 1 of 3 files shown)


nbulaj (commented):
Please do it in a backward compatible way. Maybe add a check like 

```ruby
if Doorkeeper::VERSION >= ... 
  def initiaize(...)
else
  def initialize(........)
 end
```

CircumnavigatingFlatEarther:
Travis-ci.org went offline just yesterday and we don't currently have a CI for this gem.

I have tested successfully this branch locally both with Doorkeeper >= 5.5.1 and Doorkeeper 5.5.0.

jmortlock:
Any updates on getting this merged / released? Seems to be an issue a few people are hitting and it would be nice to be able to update to the latest doorkeeper.

Antiarchitect:
Also hit this issue.

nbulaj (approved):

jhass:
Can we get a release for this please? :)

PR #155, opened 2021-06-17
Add gh actions

jmortlock:
### WHY

Switch the build from TravisCI to GitHub actions, I had to remove the build against master as GH Actions has no real equivalent + it was failing anyway.

Added Rails 6.1


jmortlock:
Had wrong base branch, need to re-pr

PR #156, opened 2021-06-17
Build with GitHub actions

jmortlock:
Switch the build over to GitHub actions, this means we lose a few things. 
 1. building against ppc64le no longer works
 2. There is no allowed_failure equivalent

Added Rails 6.1 to the build


jmortlock:
Whats the guidance here as far as built tools go? Seems you will either have to apply for travis opensource status or switch to another build system?

I would interested to here your thoughts as I would also like to contribute Ruby 3 support for this gem as well but hard when there is no test stage.

nbulaj:
Hi @jmortlock . Sorry for being so slow.
Actually there [must be a TravisCI](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/.travis.yml), but it dissipated somewhere :thinking: Looks like it happened due to planned servers shutdown.

I'm all for GH Actions, but I don't sure if they fit opensource projects with the available time per CI in month. I have to check their plans

About Ruby 3 - nice to have it :+1:

jmortlock:
If you want to continue using TravisCI you need to switch to travis.com and and apply for open source status.
```
Free for open source. We love the Open Source Community, and to show how much we love it, upon validated request placed with our Support Team you may receive free OSS credits for your public builds. 
```

As far as I can tell GitHub actions is free (currently) for open source projects and does not have any quota, I've also noticed a lot of ruby projects have made the switch from Travis to Github.

nbulaj:
Actually as we moved most of the Doorkeeper repos under GH Actions this one should be move as well.

@jmortlock do you still want to update this PR so we can later merge it? Thanks!

jmortlock:
Hi @nbulaj I have updated this PR and followed the CI file from the main doorkeeper Gem

I am keen to get this merged and than align version support with what you have in the main project, namely ruby 3+ and Rails 7

Thanks

nbulaj (approved):

PR #167, opened 2022-07-01
`introspection_endpoint` cannot be defined when token introspection is turned off

### LINKED ISSUE #166: `/.well-known/openid-configuration` crashes when `Doorkeeper.config.allow_token_introspection` is false
Dookeeper allows to disable token introspection by setting `allow_token_introspection` to `false`. With this configured, the route is not drawn and naturally `oauth_introspect_url` helper is not available.

https://github.com/doorkeeper-gem/doorkeeper/blob/v5.6.0.rc1/lib/doorkeeper/rails/routes.rb#L39

However, on the doorkeeper-openid_connect side, the above behavior is overlooked and `oauth_introspect_url` is always depended on, which causes NoMethodError whenever we configure as such and make request to `/.well-known/openid-configuration`.

This is reproducible if you just write `allow_token_introspection false` in `spec/dummy/config/initializers/doorkeeper.rb` and run tests.

sato11:
This fixes #166.

Since `oauth_introspect_url` is not available, let's workaround by checking if it can be responded to before actually calling it. Another possible solution is to replicate the same condition as doorkeeper repo, namely `Doorkeeper.config.allow_token_introspection.is_a?(FalseClass)`, which I've not chosen because I thought it'd make it more complex to depend on the configuration of a different gem.

Getting rid of `introspection_endpoint` field can be justified because it is a non-standard (as described in #72, the original PR which introduced it) and not required nor defined by [the specification of OIDC Discovery](https://openid.net/specs/openid-connect-discovery-1_0.html).

About the test case added, the entire routes must be reloaded because the configuration here is loaded once and routes drawn once, refusing to be overriden just like this:

```ruby
before { Doorkeeper.configure { allow_token_introspection false } }
```

So I've chosen to mock it and reload routes before and after the example case, knowing this is perhaps a little clumsy type. What do you think?

### DIFF
--- a/app/controllers/doorkeeper/openid_connect/discovery_controller.rb
+++ b/app/controllers/doorkeeper/openid_connect/discovery_controller.rb
@@ -30,7 +30,7 @@ module Doorkeeper
           authorization_endpoint: oauth_authorization_url(authorization_url_options),
           token_endpoint: oauth_token_url(token_url_options),
           revocation_endpoint: oauth_revoke_url(revocation_url_options),
-          introspection_endpoint: oauth_introspect_url(introspection_url_options),
+          introspection_endpoint: respond_to?(:oauth_introspect_url) ? oauth_introspect_url(introspection_url_options) : nil,
           userinfo_endpoint: oauth_userinfo_url(userinfo_url_options),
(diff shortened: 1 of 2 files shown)


nbulaj (approved):
Also LGTM :+1:

PR #172, opened 2022-07-11
Webfinger now returns issuer location

### LINKED ISSUE #171: Using `root_url` in `#webfinger_response` can violate specification
The documentation says in [2. OpenID Provider Issuer Discovery](https://openid.net/specs/openid-connect-discovery-1_0.html#IssuerDiscovery):

> The Issuer location MUST be returned in the WebFinger response as the value of the href member of a links array element with rel member value http://openid.net/specs/connect/1.0/issuer.

which then becomes the premise of [4. Obtaining OpenID Provider Configuration Information](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig):

>OpenID Providers supporting Discovery MUST make a JSON document available at the path formed by concatenating the string /.well-known/openid-configuration to the Issuer.

It seems it is assumed that our `href` must be exactly what our `issuer` configuration points to. And since `issuer` can be configured so that it does not always be the `root_url`, I feel `#webfinger_response` should return `issuer` value and not the `root_url`.

For example, when `issuer` is configured like this;

```ruby
  issuer 'https://example.com/issuer1'
end
```

the response should be like this, provided that we have provider configuration available at `https://example.com/issuer1/.well-known/openid-configuration`.

```json
{
  "subject": "test@example.com",
  "links": [
    {
      "rel": "http://openid.net/specs/connect/1.0/issuer",
      "href": "https://example.com/issuer1"
    }
  ]
}
```

sato11:
Also removes the definition of `webfinger_url_options`, as it now becomes irrelevant.

This closes #171. Detailed background is provided in the issue.

Removing configurability of `webfinger_url_options` can be non-trivial. However if we are going to let it stick around, some might feel that another option like `provider_url_options` is missing. I wouldn't prefer adding it, because these `/.well-known` stuff are supposed to be self-sufficient, unlike endpoints that `#provider` might return. How does that sound?

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ The following settings are optional:
 - `discovery_url_options`
   - The URL options for every available endpoint to use when generating the endpoint URL in the
     discovery response. Available endpoints: `authorization`, `token`, `revocation`,
-    `introspection`, `userinfo`, `jwks`, `webfinger`.
+    `introspection`, `userinfo`, `jwks`.
   - This option requires option keys with an available endpoint and
     [URL options](https://api.rubyonrails.org/v6.0.3.3/classes/ActionDispatch/Routing/UrlFor.html#method-i-url_for)
     as value.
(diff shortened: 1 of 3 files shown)


nbulaj:
Hello @sato11 . If you're still interested - can you please check above and rebase the PR? :pray:  Thanks!

nbulaj:
Closing in favor of #250 , thanks! :bow:

PR #175, opened 2022-08-23
Align support with doorkeeper

jmortlock:
Update the test suite and made the following changes

- Add Ruby 3.0 and Ruby 3.1
- Add Rails 7.0
- Drop Ruby 2.5
- Removed shims used in test mode to allow compatibility with older rubies and rails versions.

This brings it inline with whats tested in the main doorkeeper gem


jmortlock:
@nbulaj  Anything else you want from this PR ?

nbulaj (approved):
LGTM :+1:

nbulaj:
Thanks @jmortlock , nothing more for this PR :bow:

PR #176, opened 2022-09-14
Update json-jwt version for OpenSSL 3

ShalaQweghen:
The `json-jwt` version used by the current gem is [incompatible with OpenSSL 3.0](https://github.com/nov/json-jwt/pull/102). This is especially a problem with operating systems that are only shipped with OpenSSL 3.0 like Ubuntu 22.04 LTS.


nbulaj (approved):
LGTM :+1: 
Thanks!

PR #177, opened 2022-11-02
feat: remove json-jwt, migrate ruby-jwt

kristof-mattei:


phlegx:
@kristof-mattei this was my try time ago: https://github.com/phlegx/doorkeeper-openid_connect/tree/feature/change-to-jwt-gem

kristof-mattei:
@phlegx ha, seems like I could've saved myself some time.

Couple of things: 

1) I used the same thumbprint as the `json-jwt` gem, so less changes
2) No need for the `normalized` method to slice the data, the `export` filters that all out for you

phlegx:
@kristof-mattei very nice! So we dont need to slice anymore. Great job.

phlegx:
@toupeira looks good. What do you think? Time to move to `ruby-jwt`.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 - [#] Add here
 
+## Unreleased next
+
+- [#???] Replace `json-jwt` with `ruby-jwt` to align with doorkeeper-jwt.
```
toupeira, line 7: **nitpick:** We can use the PR itself here:

```suggestion
- [#177] Replace `json-jwt` with `ruby-jwt` to align with doorkeeper-jwt.
```

toupeira:
@kristof-mattei thanks for this contribution, and apologies for the delay!

This LGTM, but I'm not really active in this project anymore and currently even lack a dev setup to test out these changes in a real app :sweat_smile: I don't really expect any problems though.

@nbulaj could you give this a look too? :pray:

nbulaj (approved):
LGTM :+1: 

Just changelog entry is missed.
Hope it won't break existing app setups

kristof-mattei:
> LGTM :+1: 
> 
> 
> 
> Just changelog entry is missed.
> 
> Hope it won't break existing app setups

I'll update the changelog later today. I'll also add the changes of some string to symbol. 

In terms of impacting other apps: does this warrant a major bump?

toupeira:
> In terms of impacting other apps: does this warrant a major bump?

@kristof-mattei @nbulaj probably not since nothing *should* break (or even change), but one method I used in the past is to cut a [prerelease gem](https://guides.rubygems.org/patterns/#prerelease-gems), and then do a proper release after a few weeks, or once a user confirmed it to be working.

nbulaj:
> Do we consider the JWK object as public API? Or just the JSON output. If the former:

The problem could be if somebody relied on `json-jwt` under the hood (and done some configuring for it, or patched, or whatever). It;s super common case actually. Just like with Doorkeeper internals - I saw a lot of cases when somebody patched them and then we break some private API :smiley: 

Anyway I think it's OK to bump a minor (not patch) version as we actually don't break public API, yes, so I hope upgrading the gem should be smooth. Thanks @kristof-mattei

PR #181, opened 2023-01-11
Unpin ruby version

sato11:
Hi 👋 I find the existence of .ruby-version keeping me from randomly using with Ruby 3.2.

There has been one since [the initial commit](https://github.com/jmortlock/doorkeeper-openid_connect/commit/7dfc9bbce77b22bed5a3fe39350236835bc2c98a#diff-1bdb3279c7718f7037156f83c80eb81c58d37b286e027219e055b76249082264) but I bet it is unconventional to include it in gem repository. For example it disables us from `bundle install` with any versions of Ruby but the one specified, which I reckon is disadvantageous.


nbulaj (approved):
Yeah, agree, better not to have it. We should allow any local Ruby version for development :+1:

PR #182, opened 2023-01-11
Bump the range of supported Ruby and Rails versions

sato11:
Ruby 3.2 is out and 2.6 has reached the EOL:
https://www.ruby-lang.org/en/downloads/branches/

Also since Rails 5 has been dropped support for and
corresponding gemfiles are now dispensable to ci.

This is a follow-up to what I have committed to doorkeeper gem in https://github.com/doorkeeper-gem/doorkeeper/pull/1618.  I guess it would be great if the CI setup mirrors what doorkeeper has just like #175 has achieved it.

As 2.6 is dropped from CI I'm including the change that bumps the required ruby version. What do you think?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,23 +21,16 @@ jobs:
       matrix:
         os: [ ubuntu-latest ]
         ruby:
-          - 2.6
           - 2.7
           - '3.0'
           - '3.1'
+          - '3.2'
           - head
         gemfile:
-          - gemfiles/rails_5.0.gemfile
-          - gemfiles/rails_5.2.gemfile
           - gemfiles/rails_6.0.gemfile
           - gemfiles/rails_6.1.gemfile
           - gemfiles/rails_7.0.gemfile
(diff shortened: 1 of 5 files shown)


sato11:
Forgot to mention it in the changelog. Added it 🙂

nbulaj:
@sato11 could you please fix the conflicts? :pray: Thanks!

.github/workflows/ci.yml
```diff
@@ -36,8 +36,6 @@ jobs:
         exclude:
           - ruby: 3.0
             gemfile: gemfiles/rails_5_2.gemfile
```
nbulaj, line 38: We can also drop Rails 5.0 and 5.2

sato11: Right. Will commit it here since you mention it 👍

sato11: @nbulaj How does it look to you now? 🙂

nbulaj (approved):
LGTM :+1:

PR #183, opened 2023-01-17
stop render consent screen when user is not logged-in

nov:
current `doorkeeper-openid_connect` implementation always renders consent screen when `prompt=consent` is requested.
however, user can be **_non-authenticated_** status when requesting.
in that case, application tend to redirect to login page, and it conflicts with rendering consent screen, and results in `AbstractController::DoubleRenderError` error.

this pull request stop rendering consent screen when user is not authenticated, and will render consent screen after the user is authenticated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - [#PR ID] Add your changelog entry here.
+- [#183] stop render consent screen when user is not logged-in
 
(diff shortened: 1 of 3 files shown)


nbulaj (approved):
LGTM :+1: 

Can we add a changelog entry for this one? :pray:

nov:
ok, added changelog.

nbulaj:
Thanks!

PR #188, opened 2023-02-01
Fix dookeeper-jwt compatibility

### LINKED ISSUE #187: uninitialized constant Doorkeeper::JWT::JWK in 1.8.4
I just upgraded to `doorkeeper-openid_connect` version 1.8.4 and my application is no longer running with the following error and stack trace:

```
JWT::JWK.new(key)
       ^^^^^> with backtrace:

# /Users/user/.gem/ruby/3.1.3/gems/doorkeeper-openid_connect-1.8.4/lib/doorkeeper/openid_connect.rb:51:in `signing_key'
# /Users/user/.gem/ruby/3.1.3/gems/doorkeeper-openid_connect-1.8.4/lib/doorkeeper/openid_connect/id_token.rb:35:in `as_jws_token'
```

zavan:
Using this gem along with the `doorkeeper-jwt` gem caused problems because that gem defines a `Doorkeeper::JWT` constant, so the autoloader thinks it should load `JWT::JWK` and other constants under the `JWT` namespace from there since this gem is also under the `Doorkeeper` namespace. The solution is to prefix the `JWT` uses with `::`.

Fixes #187.

### DIFF
--- a/lib/doorkeeper/openid_connect.rb
+++ b/lib/doorkeeper/openid_connect.rb
@@ -48,7 +48,7 @@ module Doorkeeper
         else
           OpenSSL::PKey.read(configuration.signing_key)
         end
-      JWT::JWK.new(key)
+      ::JWT::JWK.new(key)
     end
 
     def self.signing_key_normalized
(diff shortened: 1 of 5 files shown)


zavan:
Fixes #187.

nbulaj (approved):
Stupid error :smile: 
Thanks!

PR #194, opened 2023-04-13
Default to RFC 7638 kid fingerprint generation

### LINKED ISSUE #193: `kid` value in headers in different format after upgrading from 1.8.3 to 1.8.5
This may not be an issue for everyone but I wanted to post an issue warning to others who are upgrading or might be facing an issue with the upgrade as this upgrade (`1.8.3` -> `1.8.5`) caused an authentication incident for my team today.

##  Source of the problem

The root of the issue is this pull request: https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/177

This was [called out as a potential issue](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/177#issuecomment-1338902262) in the PR itself by @nbulaj .

Specifically, [this line of code](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/01903c81a2b6237a3bf576ed45864f69ef20184e/lib/doorkeeper/openid_connect/id_token.rb#L37).

The OpenID kid is an opaque value that is supposed to identify public keys. `json-jwt` and `ruby-jwt` implement the kid slightly differently:

[`kid` format for json-jwt](https://github.com/nov/json-jwt/blob/413848c8c93a99093cbb68f198004848f62ecdbd/lib/json/jwk.rb#L28-L38) (used by `1.8.3`)

[`kid` format for ruby-jwt ](https://github.com/jwt/ruby-jwt/blob/18abb5c6e9e6a84cb7793be9a955f9ff1c4505ee/lib/jwt/jwk/hmac.rb#L57-L61)(used by `1.8.5`)

Normally `doorkeeper-openid_connect` manages both the [generation of an OpenID token](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/9de10c00fcd0b40f0f0a381b860501e566ee87e2/lib/doorkeeper/openid_connect/id_token.rb#L37) and the [endpoint that allows clients to discover the public keys](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/65526591251549456641f969cb8298a43c66cbd0/app/controllers/doorkeeper/openid_connect/discovery_controller.rb#L104-L115).

In our application, however, we override `Doorkeeper::OpenidConnect::DiscoveryController`  and for the JWK output in our controller we use call `to_jwk`, which uses `json-jwt` NOT `ruby-jwt`

Because our override of `Doorkeeper::OpenidConnect::DiscoveryController`  assumes that the `kid` format will be the format used by `json-jwt`, an upgrade to  doorkeeper-openid_connect`1.8.5` resulted in an incident for us because any service that [validates the signature of a JWT using the `kid` header value](https://self-issued.info/docs/draft-jones-json-web-token-01.html#ReservedHeaderParameterName) against the `kid` in our keys discovery endpoint will raise an error (because the `kid` s do not match).

## What I actually see 

Screenshot of a (test) decoded JWT generated with `1.8.5`:

## What I expect to see

The `kid` should match the `kid` value of my JWK#kid in my discovery keys endpoint (`"OcGcrSBrUtT8mNNNkm1eofXGmOgM4SQtqFgFT8BCszU"`)
Screenshot of a (test) decoded JWT generated with `1.8.3`:

stanhu:
The switch from the `json-jwt` to `jwt` gem in #177 changed the default `kid` generation from RFC 7638
(https://www.rfc-editor.org/rfc/rfc7638) to a format based on the SHA256 digest of the key elements.

However, clients may fail if the the `kid` generated by `IdToken` does not match a key listed in JWKS discovery endpoint, which may be implemented by the application using RFC 7638-based `kid` values. To restore the previous behavior, applications have to set a global setting:

```
JWT.configuration.jwk.kid_generator_type = :rfc7638_thumbprint
```

However, relying on this global setting is not ideal since other keys may depend on the legacy `kid` values.

In keeping with semantic versioning, restore the `kid` generation to RFC 7638. Whether this should be customizable can be discussed later.

Closes #193

### DIFF
--- a/lib/doorkeeper/openid_connect.rb
+++ b/lib/doorkeeper/openid_connect.rb
@@ -48,7 +48,7 @@ module Doorkeeper
         else
           OpenSSL::PKey.read(configuration.signing_key)
         end
-      ::JWT::JWK.new(key)
+      ::JWT::JWK.new(key, { kid_generator: JWT::JWK::Thumbprint })
     end
 
     def self.signing_key_normalized
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - [#PR ID] Add your changelog entry here.
+- [#194] Default to RFC 7638 kid fingerprint generation (thanks to @stanhu).
 
 ## v1.8.5 (2023-02-02)
 
@@ -9,6 +10,9 @@
 - [#188] Fix dookeeper-jwt compatibility (thanks to @zavan).
 
 ## v1.8.4 (2023-02-01)
+
+Note that v1.8.4 changed the default kid fingerprint generation from RFC 7638 to a format
+based on the SHA256 digest of the key element. To restore the previous behavior, upgrade to v1.8.6.
 
 - [#177] Replace `json-jwt` with `ruby-jwt` to align with doorkeeper-jwt (thanks to @kristof-mattei).
 - [#185] Don't call active_record_options for Doorkeeper >= 5.6.3 (thanks to @zavan).
--- a/spec/dummy/config/initializers/jwt.rb
+++ b/spec/dummy/config/initializers/jwt.rb
@@ -1,3 +1,0 @@
-# frozen_string_literal: true
-
-::JWT.configuration.jwk.kid_generator_type = :rfc7638_thumbprint
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -53,7 +53,6 @@ RSpec.configure do |config|
 
   # Reinitialize configuration after each example
   config.after do
-    load Rails.root.join('config/initializers/jwt.rb')
     load Rails.root.join('config/initializers/doorkeeper.rb')
     load Rails.root.join('config/initializers/doorkeeper_openid_connect.rb')
   end


stanhu:
Failing tests will be fixed by https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/195.

lib/doorkeeper/openid_connect.rb
```diff
@@ -48,7 +48,7 @@ def self.signing_key
         else
           OpenSSL::PKey.read(configuration.signing_key)
         end
-      ::JWT::JWK.new(key)
+      ::JWT::JWK.new(key, { kid_generator: JWT::JWK::Thumbprint })
```
jessieay, line 51: **Suggestion**: add this change to the [unreleased section of the Changelog file ](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/CHANGELOG.md#unreleased) and maybe backfill some info into the `1.8.4` Changelog so that folks upgrading to that version know about the change in `kid` format for that version?

jessieay:
Thank you for the detailed explanation on these `kid` formats @stanhu !  TIL `RFC 7638` exists :)

lib/doorkeeper/openid_connect.rb
nbulaj: +1 @stanhu :point_up:

stanhu: @nbulaj Done!

nbulaj (approved):
LGTM :+1:

nbulaj:
Released with 1.8.6

PR #201, opened 2023-11-29
Add back typ=JWT to header

Zeko369:
Along with #193, upgrade from 1.8.3 -> 1.8.5 caused and incident for us because of the underlying library changing and not having `typ: 'JWT'` in the header any more

While that's an optional part of the JWT spec, some client libraries are expecting that so I propose we add it back

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@
 /pkg
 gemfiles/*.lock
 *.gem
+
+.idea/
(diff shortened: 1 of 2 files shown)


nbulaj:
Hey @Zeko369 . This looks OK to me. Can we add a changelog entry for this PR?

nbulaj:
Will do myself

PR #204, opened 2024-01-25
Update for doorkeeper 5.6.8

### LINKED ISSUE #203: Doorkeeper patch v5.6.8 modified exception handling to require objects instead of symbols
Doorkeeper now expects objects that respond to `name_for_response` rather than just a symbol describing why the exception occurred.

Ref https://github.com/doorkeeper-gem/doorkeeper/issues/1685

bockets:
Doorkeeper patch version 5.6.8 replaced using symbols for errors with objects.

Closes #203

### DIFF
--- a/lib/doorkeeper/oauth/id_token_request.rb
+++ b/lib/doorkeeper/oauth/id_token_request.rb
@@ -21,7 +21,7 @@ module Doorkeeper
       end
 
       def deny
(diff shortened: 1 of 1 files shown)


bockets:
This will be fixed with a better solution in Doorkeeper 5.6.9. Ref https://github.com/doorkeeper-gem/doorkeeper/issues/1685

PR #209, opened 2024-07-26
Configuration per IdToken expiration

martinezcoder:
In some use cases we need to specify a different expiration time for specific tokens, e.g. in our company we create tokens for admins who can access the applications as if they were the customer by using what we call "impersonated sessions". In this scenario we want shorter IdTokens.

I imagine there could be multiple scenarios where we don't want to use the global expiration time configured. I hope this will help.

### DIFF
--- a/lib/doorkeeper/openid_connect/id_token.rb
+++ b/lib/doorkeeper/openid_connect/id_token.rb
@@ -7,11 +7,12 @@ module Doorkeeper
 
       attr_reader :nonce
 
-      def initialize(access_token, nonce = nil)
+      def initialize(access_token, nonce = nil, expires_in = Doorkeeper::OpenidConnect.configuration.expiration)
         @access_token = access_token
         @nonce = nonce
         @resource_owner = Doorkeeper::OpenidConnect.configuration.resource_owner_from_access_token.call(access_token)
         @issued_at = Time.zone.now
+        @expires_in = expires_in
       end
 
       def claims
@@ -57,7 +58,7 @@ module Doorkeeper
       end
 
       def expiration
-        (@issued_at.utc + Doorkeeper::OpenidConnect.configuration.expiration).to_i
+        (@issued_at.utc + @expires_in).to_i
       end
 
       def issued_at
--- a/spec/lib/id_token_spec.rb
+++ b/spec/lib/id_token_spec.rb
@@ -34,6 +34,16 @@ describe Doorkeeper::OpenidConnect::IdToken do
       )
     end
 
+    context 'when expires_in is specified for the token' do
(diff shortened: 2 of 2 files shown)


nbulaj (approved):
LGTM. Thanks!

phlegx:
Can you show an example please, how to use this possibility?

martinezcoder:
> Can you show an example please, how to use this possibility?

@phlegx yes, sure. You can do it by creating your own token generator class:

```ruby
  class IdTokenGenerator
    attr_reader :user_id, :client_application_id, :expires_in

    def initialize(
      user_id:,
      client_application_id:,
      expires_in: 10.minutes
    )
      @user_id = user_id
      @client_application_id = client_application_id
      @expires_in = expires_in
    end

    def id_token
      @id_token ||= Doorkeeper::OpenidConnect::IdToken.new(access_token, nil, @expires_in)
    end

    delegate :as_jws_token, to: :id_token, prefix: true

    def access_token
      @access_token ||= Doorkeeper::AccessToken.create!(
        application_id: client_application_id,
        resource_owner_id: user_id,
        expires_in: expires_in
      )
    end
  end

```

jurruh:
@martinezcoder How do we configure the `IdTokenGenerator` in your example? .new is directly called on `IdToken`: https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/3ed61241b79b19b1474a27b55fdfcb18d0f1caaa/lib/doorkeeper/oauth/id_token_request.rb#L31



@nbulaj Is there a reason that we can't implement the expiration like the other configuration options with a block where the resource owner and application are passed to?

Something like this:
```
  expiration do |resource_owner, application|
    application.expiration
  end
```

I can create a PR for this if you want.

martinezcoder:
> @martinezcoder How do we configure the `IdTokenGenerator` in your example? .new is directly called on `IdToken`:
> 
> https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/3ed61241b79b19b1474a27b55fdfcb18d0f1caaa/lib/doorkeeper/oauth/id_token_request.rb#L31
> 
> @nbulaj Is there a reason that we can't implement the expiration like the other configuration options with a block where the resource owner and application are passed to?
> 
> Something like this:
> 
> ```
>   expiration do |resource_owner, application|
>     application.expiration
>   end
> ```
> 
> I can create a PR for this if you want.

@jurruh but this is already implemented. The code I shared is only a custom solution I have to create an id_token with custom expiration time. 

The class itself is using the configured option that already exists. See that the initializer injects the default configured option:

```ruby
def initialize(access_token, nonce = nil, expires_in = Doorkeeper::OpenidConnect.configuration.expiration)
```

PR #211, opened 2024-10-14
Fix DiscoveryController#provider_response for Ruby 3.0

eduresende:
This PR aims to fix the warning message on DiscoveryController:

```
Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

These methods expect keword arguments instead of hashes, so we need to use the `**` operator.


nbulaj:
hey @eduresende .

> These methods expect keword arguments instead of hashes, so we need to use the ** operator.

I don't remember Rails route helpers ever required kwargs TBH :thinking: Can you please point to some docs? Also will be great to rebase this branch to have all the latest fixes in

eduresende:
Not necessary

PR #214, opened 2024-11-12
ci: make all tests pass again

stanhu:
The tests were failing because the current gemspec allows sqlite3 >= 1.3.6, but the external Gemfiles used older versions.

To avoid this, BUNDLE_GEMFILE needs to be set as documented in https://github.com/ruby/setup-ruby:

To use a Gemfile which is not at the root or has a different name, set BUNDLE_GEMFILE in the env at the job level as shown in the example.

### DIFF
--- a/doorkeeper-openid_connect.gemspec
+++ b/doorkeeper-openid_connect.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.7'
 
-  spec.add_runtime_dependency 'doorkeeper', '>= 5.5', '< 5.8'
+  spec.add_runtime_dependency 'doorkeeper', '>= 5.5', '< 5.9'
   spec.add_runtime_dependency 'jwt', '>= 2.5'
 
   spec.add_development_dependency 'conventional-changelog', '~> 1.2'
--- a/doorkeeper-openid_connect.gemspec
+++ b/doorkeeper-openid_connect.gemspec
@@ -27,8 +27,11 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency 'doorkeeper', '>= 5.5', '< 5.9'
   spec.add_runtime_dependency 'jwt', '>= 2.5'
 
+  spec.add_development_dependency 'bigdecimal'
   spec.add_development_dependency 'conventional-changelog', '~> 1.2'
+  spec.add_development_dependency 'drb'
   spec.add_development_dependency 'factory_bot'
+  spec.add_development_dependency 'mutex_m'
   spec.add_development_dependency 'pry-byebug'
   spec.add_development_dependency 'rspec-rails'
   spec.add_development_dependency 'sqlite3', '>= 1.3.6'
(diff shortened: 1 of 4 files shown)


stanhu:
@toupeira Ok, this should looks like it's all green now. 😄

stanhu:
Spoke too soon. Looks like `mutex_m` is being moved to a bundled gem in Ruby head: https://bugs.ruby-lang.org/issues/19885

grzesiek:
+1

stanhu:
@toupeira Could you take a look at this? We haven't heard from @nbulaj for a while. I'm also happy to take on maintainership duties if you want to add me to this project and Rubygems.

nbulaj:
Hey @stanhu   . I see all your hard-work for this project, but unfortunately I’m in a hospital with pneumonia so I’ll be able to check all these later

stanhu:
@nbulaj Oh, no, I'm sorry to hear that! I hope you get better soon.

spec/dummy/config/environments/test.rb
```diff
@@ -19,7 +19,10 @@
   config.action_controller.perform_caching = false
 
   # Raise exceptions instead of rendering exception templates.
-  config.action_dispatch.show_exceptions = false
+  # Rails 7.1 deprecated false in favor of :none, but we need to use false for
+  # backwards compatibility: https://github.com/rails/rails/pull/45867
+  config.action_dispatch.show_exceptions =
+    Gem::Version.new(Rails.version) >= Gem::Version.new('7.1.0') ? :none : false
```
tachyons, line 25:
```suggestion
    Rails.gem_version >= Gem::Version.new('7.1.0') ? :none : false
```

Syntax sugar

doorkeeper-openid_connect.gemspec
```diff
@@ -24,11 +24,14 @@ Gem::Specification.new do |spec|
...
 
-  spec.add_runtime_dependency 'doorkeeper', '>= 5.5', '< 5.8'
+  spec.add_runtime_dependency 'doorkeeper', '>= 5.5', '< 5.9'
   spec.add_runtime_dependency 'jwt', '>= 2.5'
 
+  spec.add_development_dependency 'bigdecimal'
   spec.add_development_dependency 'conventional-changelog', '~> 1.2'
+  spec.add_development_dependency 'drb'
   spec.add_development_dependency 'factory_bot'
+  spec.add_development_dependency 'mutex_m'
```
nbulaj, line 34: Why do we need `mutex_m` and `bigdecimal`? :thinking:

stanhu: @nbulaj Ruby 3.4 moved these gems out: https://docs.ruby-lang.org/en/master/NEWS_md.html

nbulaj: Wow, didn't know! Thanks :bow:

nbulaj (approved):
Thanks!

stanhu:
@nbulaj Hope you are feeling better. Thanks for the merge!

When you get a chance, could you tag a new release? Thank you!

PR #216, opened 2024-12-04
Test against Rails 7.1, 7.2, 8.0

naitoh:
Rails 7.1, 7.2, 8.0 are out
corresponding gemfiles are now dispensable to ci.
https://rubyonrails.org/maintenance

This is a follow-up to what I have committed to doorkeeper gem in https://github.com/doorkeeper-gem/doorkeeper/pull/1752.
I guess it would be great if the CI setup mirrors what doorkeeper has just like https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/182 has achieved it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,15 @@ jobs:
           - head
         gemfile:
           - gemfiles/rails_7.0.gemfile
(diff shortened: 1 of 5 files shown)


nbulaj (approved):
LGTM :bow:

PR #215, opened 2024-11-21
Bump the range of supported Ruby and Rails versions

naitoh:
Ruby 3.3 is out and 2.7, 3.0 has reached the EOL:
https://www.ruby-lang.org/en/downloads/branches/

Also since Rails 6 has been dropped support for and corresponding gemfiles are now dispensable to ci.
https://rubyonrails.org/maintenance

PR #219, opened 2025-01-17
Bump the range of supported Ruby versions

naitoh:
### 1. concurrent-ruby 1.3.5 don't work on Rails 7.0

> NameError:
>  uninitialized constant ActiveSupport::LoggerThreadSafeLevel::Logger

See: https://github.com/ruby-concurrency/concurrent-ruby/pull/1062

### 2. concurrent-ruby 1.3.4 don't work on Ruby head

> LoadError:
> cannot load such file -- logger

Exclude the combination of Ruby head and Rails 7.0

This is a follow-up to what I have committed to doorkeeper gem in https://github.com/doorkeeper-gem/doorkeeper/pull/1760.
I guess it would be great if the CI setup mirrors what doorkeeper has just like https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/215 has achieved it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,7 @@ jobs:
           - '3.1'
           - '3.2'
           - '3.3'
+          - '3.4'
           - head
         gemfile:
           - gemfiles/rails_7.0.gemfile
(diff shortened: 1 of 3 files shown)


nbulaj (approved):
Thanks!

PR #220, opened 2025-01-31
make prompt priority static

nov:
current implementation behaves differently on these patterns
* `prompt=select_account+login`
* `prompt=login+select_account`

since OIDC spec doesn't mention about prompt value order matter or not, it's totally up to implementation.
however, in most cases, developer will expect this behavior regardless of the order.

1. `prompt=login` clears current authenticated session
2. then `prompt=select_account` tries to render account selector
3. if account selector cannot list-up accounts, show login page etc
    * this is application specific behaviour, not related to this gem

so, this pull request is trying to make the prompt handling order static based on the priority of each prompt value.

### DIFF
--- a/spec/controllers/doorkeeper/authorizations_controller_spec.rb
+++ b/spec/controllers/doorkeeper/authorizations_controller_spec.rb
@@ -307,9 +307,8 @@ describe Doorkeeper::AuthorizationsController, type: :controller do
 
       # FIXME:
       it 'when login+consent' do
-        expect do
-          authorize! prompt: 'login consent'
-        end.to raise_error AbstractController::DoubleRenderError
+        authorize! prompt: 'login consent'
+        expect(response).to redirect_to('/reauthenticate')
       end
     end
 
--- a/spec/controllers/doorkeeper/authorizations_controller_spec.rb
+++ b/spec/controllers/doorkeeper/authorizations_controller_spec.rb
@@ -284,6 +284,35 @@ describe Doorkeeper::AuthorizationsController, type: :controller do
       end
     end
 
+    context 'with multiple prompt values' do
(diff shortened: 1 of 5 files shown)


nov:
spec seems need this update
https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/219

spec/controllers/doorkeeper/authorizations_controller_spec.rb
```diff
@@ -284,6 +284,35 @@ def expect_successful_callback!
...
+        authorize! prompt: 'select_account consent'
+        expect(response).to redirect_to('/select_account')
+      end
+
+      # FIXME:
+      it 'when login+consent' do
+        expect do
+          authorize! prompt: 'login consent'
+        end.to raise_error AbstractController::DoubleRenderError
+      end
```
nov, line 313: this is existing bug, but not trying to fix it in this pull request. just making the fact visible.

nbulaj (approved):
Nice, thanks!

nbulaj:
#219 was merged so we can rebase I believe, thanks! :bow:

spec/controllers/doorkeeper/authorizations_controller_spec.rb
nov: tried to fix, and seems succeeded?

nov:
hum, another error on CI...

nov:
OK, CI needs one more fix
https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/223

nbulaj:
Can we please also add a changelog entry? :pray:

nov:
OK, added changelog.

nbulaj:
Thanks!

PR #221, opened 2025-02-02
allow prompt=create use, but not support in this gem

nov:
to enable using this extension by RP developers, just stop raising error.
https://openid.net/specs/openid-connect-prompt-create-1_0.html

no need to support `create` value in this gem itself.

### DIFF
--- a/lib/doorkeeper/openid_connect/helpers/controller.rb
+++ b/lib/doorkeeper/openid_connect/helpers/controller.rb
@@ -79,6 +79,8 @@ module Doorkeeper
               render :new if owner
             when 'select_account'
               select_account_for_oidc_resource_owner(owner)
+            when 'create'
+              # NOTE: not supported, but not raise error.
             else
               raise Errors::InvalidRequest
             end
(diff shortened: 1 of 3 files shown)


lib/doorkeeper/openid_connect/helpers/controller.rb
```diff
@@ -84,6 +84,8 @@ def handle_oidc_prompt_param!(owner)
               render :new if owner
             when 'select_account'
               select_account_for_oidc_resource_owner(owner)
+            when 'create'
+              # NOTE: not supported, but not raise error.
             else
               raise Errors::InvalidRequest
             end
```
nov, line 91: it might be better not to raise `invalid_request` errors on any unknown prompt values.

> If an OP receives a prompt value outside the set defined above that it does not understand, it MAY return an error or it MAY ignore it; in practice, not returning errors for not-understood values will help facilitate phasing in extensions using new prompt values.
https://openid.net/specs/openid-connect-core-1_0.html

nbulaj (approved):
LGTM, thanks!

PR #227, opened 2025-09-18
Add support for configurable registration endpoint in OpenID Connect discovery

davidraj:
- Updated CHANGELOG to reflect the new feature.
- Enhanced README with details on the `registration_endpoint` configuration.
- Implemented logic in the discovery controller to include the registration endpoint if configured.
- Added configuration option for `registration_endpoint` in the Doorkeeper OpenID Connect settings.
- Created tests to verify the behavior of the registration endpoint in the discovery response.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- Add support for configurable registration_endpoint in OpenID Connect discovery
 - [#PR ID] Add your changelog entry here.
 
 ## v1.8.11 (2025-02-10)
--- a/README.md
+++ b/README.md
@@ -163,6 +163,17 @@ The following settings are optional:
   - Used by implementations like https://github.com/IdentityModel/oidc-client-js.
   - The block is executed in the controller's scope, so you have access to your route helpers.
 
+- `registration_endpoint`
+  - The URL for client registration endpoint, used for dynamic client registration.
+  - The block is executed in the controller's scope, so you have access to your route helpers.
+  - If not configured, the field will be omitted from the discovery document.
+
+  ```ruby
+  Doorkeeper::OpenidConnect.configure do
+    registration_endpoint -> { oauth_register_url }
+  end
+  ```
+
 - `discovery_url_options`
   - The URL options for every available endpoint to use when generating the endpoint URL in the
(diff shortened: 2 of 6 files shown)


szechyjs:
II agree that this is still useful, even with #230 as I would like to override the route/controller used as the out of the box registration controller functionality doesn't accomplish our requirements.

nbulaj:
Hello @szechyjs . Can't we just add our own discovery controller and add required functionality? :eyes:

55728:
Hi @szechyjs 👋 — doing a bit of PR triage here.

Totally get that the out-of-the-box registration controller doesn't fit your requirements. Just to make sure we're on the same page about the workaround @nbulaj suggested: you can run your own registration route/controller with whatever behavior you need, and then override the discovery controller so it advertises your registration_endpoint URL in .well-known. That combination should cover replacing the registration behavior end-to-end — does it unblock your case?

If the only friction is having to subclass the discovery controller just to change that one URL, we could consider a small end_session_endpoint-style lambda override for registration_endpoint as ergonomic sugar — but that's a minor convenience, so I'll leave the call on whether it's worth it to @nbulaj.

If we don't hear back in a couple of weeks I'll close this as covered by the controller-override approach. Thanks! 😊

55728:
Thank you for this PR, @davidraj.

After investigating the current `master`, the `registration_endpoint` use case is now largely covered by #230 (Dynamic Client Registration):

- `discovery_controller.rb` now conditionally includes `registration_endpoint` when DCR is enabled:
  ```ruby
  registration_endpoint: openid_connect.dynamic_client_registration ? oauth_dynamic_client_registration_url(dynamic_client_registration_url_options) : nil,
  ```
- When DCR is disabled, the value is `nil` and excluded by `.compact`, so there is no unintended endpoint exposure.
- The `dynamic_client_registration_url_options` are also generated automatically via the meta-programmed `*_url_options` methods, so `discovery_url_options` customization works out of the box.

One difference: #230 only advertises `registration_endpoint` when the gem's built-in DCR is enabled, whereas this PR allowed advertising an arbitrary (e.g. external) registration endpoint independently. If that standalone use case is still needed, we can revisit.

For now I'm closing this in favor of the DCR-based approach — thanks again for helping shape the direction here.

PR #229, opened 2025-11-09
Allow signing key and algorithm to be the proc.

Zogoo:
This will allow the application to manage private keys with more flexibility.


Zogoo:
@nbulaj, I have added the Change log. Could you please look at it?

PR #230, opened 2025-11-17
Dynamic client registration

vicocamacho:
This PR adds dynamic client registration as an optional feature that can be enabled through a configuration value.

I currently have it working in my own application and wanted to contribute it upstream so it can be refined as needed and potentially benefit other users of this gem.

Please let me know if I’ve missed anything or if you have any feedback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - [#229] Allow to application manage signing key and algorithm
+- [#230] Add dynamic client registration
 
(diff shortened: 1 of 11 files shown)


HunterHillegas:
More time required to really know for sure but have to say, I was able to take this branch and get a fully functioning DCR setup for ChatGPT working in basically no time. Seems to work great so far.

nbulaj (approved):
Small comment, otherwise LGTM from my end :muscle:

PR #232, opened 2026-02-02
feat: implements customizable OpenID request class and updates references

caiubi:
This adds a config option for the openid-request class, following same mixin principles of the base Doorkeeper gem. 

I had an issue where my app connects to multiple dbs, so I had to bypass nonce persistence completely, since I can't have this model in the default db. 

This PR addresses that issue.


nbulaj:
Yes just built a new release - [1.9.0](https://rubygems.org/gems/doorkeeper-openid_connect)

PR #233, opened 2026-02-12
fix: handle DoubleRenderError in library instead of requiring consumer workaround

### LINKED ISSUE #206: some combinations of prompt causes double rendering/redirection
`prompt=select_account+consent` & `prompt=select_account+login` causes it.

bvogel:
## Summary

Moves DoubleRenderError prevention from consumer callbacks into the library itself, so users no longer need to add `self.response_body = nil` / `@_response_body = nil` in their `reauthenticate_resource_owner` and `select_account_for_resource_owner` blocks.

Fixes #206

## Problem

Since v1.8.11, several scenarios cause `AbstractController::DoubleRenderError`:

1. **`max_age` + prompt**: `handle_oidc_max_age_param!` triggers `reauthenticate_oidc_resource_owner` (redirect), then `handle_oidc_prompt_param!` also renders/redirects
2. **Multiple prompt values**: e.g. `prompt=login consent` - `consent` renders `:new`, then `login` redirects via `reauthenticate_oidc_resource_owner`
3. **`max_age=0`**: Always triggers reauthentication, compounding with any prompt handler

The current workaround (documented in the [generator template](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/88f2df8ec0f934611070def307afb24ac67a2f76/lib/generators/doorkeeper/openid_connect/templates/initializer.rb#L33-L38) and [test dummy](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/spec/dummy/config/initializers/doorkeeper_openid_connect.rb)) requires every consumer to add response clearing in their callbacks. This is error-prone and undiscoverable.

## Solution

Adds a `clear_oidc_response!` helper that safely clears `response_body` before issuing a new render/redirect. Called in:

- `reauthenticate_oidc_resource_owner` - before executing the consumer's callback
- `select_account_for_oidc_resource_owner` - before executing the consumer's callback
- `handle_oidc_prompt_param!` consent branch - before `render :new`

This matches the existing pattern in `handle_oidc_error!` which already clears the response body.

## Changes

- **`lib/doorkeeper/openid_connect/helpers/controller.rb`** - add `clear_oidc_response!` and call it in the three locations above
- **`spec/dummy/config/initializers/doorkeeper_openid_connect.rb`** - remove the consumer-side workaround (no longer needed)
- **`lib/generators/doorkeeper/openid_connect/templates/initializer.rb`** - remove the DoubleRenderError workaround comments

All 174 existing tests pass without the consumer-side workaround.

### DIFF
--- a/lib/doorkeeper/openid_connect/helpers/controller.rb
+++ b/lib/doorkeeper/openid_connect/helpers/controller.rb
@@ -122,8 +122,6 @@ module Doorkeeper
         end
 
         def clear_oidc_response
-          return unless response_body
-
           self.response_body = nil
           @_response_body = nil
         end
--- a/spec/controllers/doorkeeper/authorizations_controller_spec.rb
+++ b/spec/controllers/doorkeeper/authorizations_controller_spec.rb
@@ -390,12 +390,31 @@ describe Doorkeeper::AuthorizationsController, type: :controller do
         expect(response).to redirect_to '/select_account'
       end
     end
+
+    context 'when used along with prompt=login' do
+      it 'redirects to reauthenticate without raising a DoubleRenderError' do
+        user.update! current_sign_in_at: 5.minutes.ago
+        authorize! max_age: 10, prompt: 'login'
+
+        expect(response).to redirect_to '/reauthenticate'
+      end
+    end
+
+    context 'when used along with prompt=consent' do
+      it 'redirects to reauthenticate without raising a DoubleRenderError' do
+        user.update! current_sign_in_at: 5.minutes.ago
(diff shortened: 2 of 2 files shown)


lib/doorkeeper/openid_connect/helpers/controller.rb
```diff
@@ -118,7 +121,15 @@ def return_without_oidc_prompt_param(prompt_value)
           return_to.to_s
         end
 
+        def clear_oidc_response!
```
Sokre95: I would maybe remove `!` from method name

PR #246, opened 2026-04-22
Fix `at_hash` to use correct hash algorithm based on `signing_algorithm`

### LINKED ISSUE #245: `at_hash` always uses SHA-256 regardless of `signing_algorithm` (OIDC Core spec violation)
### Summary

`IdTokenToken#at_hash` hardcodes `Digest::SHA256` for computing the access token hash, regardless of the configured `signing_algorithm`. This violates [OIDC Core 1.0 §3.1.3.6](https://openid.net/specs/openid-connect-core-1_0.html#HybridIDToken) and [§3.2.2.9](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitIDTValidation), which require the hash algorithm to match the signing algorithm:

| Signing Algorithm | Required Hash | Left-half bits |
|---|---|---|
| RS256 / ES256 / HS256 | SHA-256 | 128 |
| RS384 / ES384 / HS384 | SHA-384 | 192 |
| RS512 / ES512 / HS512 | SHA-512 | 256 |

When using HS512 or ES512, the `at_hash` in the ID Token is computed with SHA-256 instead of SHA-512, producing an incorrect value. Strict clients will reject the ID Token during validation.

### Root cause

In `lib/doorkeeper/openid_connect/id_token_token.rb`:

```ruby
def at_hash
  sha256 = Digest::SHA256.new  # <-- hardcoded
  token = @access_token.token
  hashed_token = sha256.digest(token)
  first_half = hashed_token[0...hashed_token.length / 2]
  Base64.urlsafe_encode64(first_half).tr('=', '')
end
```

The digest algorithm is always SHA-256, ignoring `Doorkeeper::OpenidConnect.signing_algorithm`.

### Steps to reproduce

1. Configure `signing_algorithm` to `:hs512`:

```ruby
  signing_algorithm :hs512
  signing_key "your-256-bit-secret-key-here-for-hmac-512"
  # ...
end
```

2. Issue an ID Token via the implicit flow (`response_type=id_token token`).

3. The `at_hash` claim in the ID Token is computed with SHA-256 (128-bit left-half) instead of SHA-512 (256-bit left-half).

4. A spec-compliant client verifying the `at_hash` will reject the ID Token.

### Expected behavior

`at_hash` should use the hash algorithm corresponding to the configured `signing_algorithm`:

```
HS512 → SHA-512, left 256 bits
HS384 → SHA-384, left 192 bits
HS256 → SHA-256, left 128 bits
(same for RS* and ES* families)
```

### Proposed fix

Select the digest algorithm based on `signing_algorithm`:

```ruby
def at_hash
  alg = Doorkeeper::OpenidConnect.signing_algorithm.to_s
  digest_class =
    case alg
    when /256$/ then Digest::SHA256
    when /384$/ then Digest::SHA384
    when /512$/ then Digest::SHA512
    else Digest::SHA256
    end

  digest = digest_class.digest(@access_token.token)
  Base64.urlsafe_encode64(digest[0, digest.length / 2]).tr('=', '')
end
```

### Environment

- Ruby 4.0.2
- Rails 8.0.5
- doorkeeper-openid_connect master

55728:
## Summary

`IdTokenToken#at_hash` hardcodes `Digest::SHA256` for computing the access token hash, regardless of the configured `signing_algorithm`. This violates [OIDC Core 1.0 §3.1.3.6](https://openid.net/specs/openid-connect-core-1_0.html#HybridIDToken) and [§3.2.2.9](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitIDTValidation).

The spec requires the hash algorithm to match the signing algorithm:

| Signing Algorithm | Required Hash | Left-half bits |
|---|---|---|
| RS256 / ES256 / HS256 | SHA-256 | 128 |
| RS384 / ES384 / HS384 | SHA-384 | 192 |
| RS512 / ES512 / HS512 | SHA-512 | 256 |

When using HS512 or ES512, the `at_hash` in the ID Token is computed with SHA-256 instead of SHA-512. Strict clients verifying the `at_hash` will reject the ID Token.

## Changes

- **`lib/doorkeeper/openid_connect/id_token_token.rb`**: Select digest algorithm dynamically based on the suffix of `signing_algorithm` (256/384/512).
- **`spec/lib/id_token_token_spec.rb`**: Add tests verifying correct `at_hash` for HS512 and ES512 signing algorithms.

## Before

```ruby
# signing_algorithm: :HS512
at_hash = "77QmUPtjPfzWtF2AnpK9RQ"  # SHA-256, 128-bit (WRONG)
```

## After

```ruby
# signing_algorithm: :HS512
at_hash = "q7nS86GgvvFaZkzALLWqJYaJIKw2wCDAVfCAsm5CrBM"  # SHA-512, 256-bit (CORRECT)
```

Closes #245

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Please add here
 - [#241] Fix NameError on doorkeeper master by deferring AR model loading in run_hooks (see [Doorkeeper PR](https://github.com/doorkeeper-gem/doorkeeper/pull/1804))
+- [#246] Fix `at_hash` to use correct hash algorithm based on `signing_algorithm`
 
 ## v1.9.0 (2026-03-16)
 
--- a/lib/doorkeeper/openid_connect/id_token_token.rb
+++ b/lib/doorkeeper/openid_connect/id_token_token.rb
@@ -24,11 +24,18 @@ module Doorkeeper
       #   access_token value with SHA-256, then take the left-most 128 bits and
       #   base64url-encode them. The at_hash value is a case-sensitive string.
       def at_hash
-        sha256 = Digest::SHA256.new
-        token = @access_token.token
-        hashed_token = sha256.digest(token)
+        hashed_token = at_hash_digest.digest(@access_token.token)
         first_half = hashed_token[0...hashed_token.length / 2]
         Base64.urlsafe_encode64(first_half).tr('=', '')
+      end
+
+      def at_hash_digest
+        case Doorkeeper::OpenidConnect.signing_algorithm.to_s
+        when /256\z/ then Digest::SHA256
+        when /384\z/ then Digest::SHA384
+        when /512\z/ then Digest::SHA512
+        else Digest::SHA256
+        end
       end
     end
   end
--- a/spec/lib/id_token_token_spec.rb
+++ b/spec/lib/id_token_token_spec.rb
@@ -33,4 +33,37 @@ describe Doorkeeper::OpenidConnect::IdTokenToken do
       })
     end
   end
+
+  describe '#at_hash' do
+    # Per OIDC Core 1.0 §3.1.3.6 / §3.2.2.9, at_hash must use the hash algorithm
+    # that matches the alg of the ID Token's JOSE header (e.g. HS512 -> SHA-512).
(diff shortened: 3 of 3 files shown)


nbulaj (approved):
LGTM for me as well 🙇  Thanks!

PR #250, opened 2026-04-23
Webfinger now returns issuer location

### LINKED ISSUE #171: Using `root_url` in `#webfinger_response` can violate specification

55728:
Closes #171. Supersedes #172.

This PR takes over the work originally submitted by @sato11 in #172,
which had gone stale with unresolved conflicts.

### From @sato11's original commit (cherry-picked):
- Return the configured `issuer` value instead of `root_url` in the WebFinger response
- Remove `webfinger_url_options` generation and related configurability
- Update specs and README accordingly

### Additional:
- Add spec for non-root issuer URL to cover the regression scenario described in #171

## Background

As described in #171, the previous implementation used `root_url` for the
WebFinger `links[].href` value, which can violate the OIDC Discovery
specification when the issuer is configured to a URL that differs from
the application root (e.g., includes a path component).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Please add here
 - [#241] Fix NameError on doorkeeper master by deferring AR model loading in run_hooks (see [Doorkeeper PR](https://github.com/doorkeeper-gem/doorkeeper/pull/1804))
 - [#246] Fix `at_hash` to use correct hash algorithm based on `signing_algorithm`
+* [#250] Return configured `issuer` instead of `root_url` in WebFinger response (thanks to @sato11 for the original work in #172)
 
 ## v1.9.0 (2026-03-16)
 
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ The following settings are optional:
 - `discovery_url_options`
   - The URL options for every available endpoint to use when generating the endpoint URL in the
     discovery response. Available endpoints: `authorization`, `token`, `revocation`,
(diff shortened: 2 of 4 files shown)


nbulaj (approved):
LGTM, thanks!

PR #254, opened 2026-04-24
Omit `expires_in` from pure `id_token` response

### LINKED ISSUE #253: Pure `id_token` response includes `expires_in` (no access token is returned)
When using the Implicit Flow with `response_type=id_token`, the response body includes `expires_in`:

```ruby
# lib/doorkeeper/oauth/id_token_response.rb
def body
  {
    expires_in: auth.token.expires_in_seconds,
    state: pre_auth.state,
    id_token: id_token.as_jws_token
  }
end
```

[OpenID Connect Core 1.0 §3.2.2.5](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthResponse) defines `expires_in` as the expiration time of the **Access Token** in seconds. In the pure `id_token` response, no access token is returned — the existing spec already verifies that `access_token` is excluded — so `expires_in` is meaningless in this context and could mislead clients into interpreting it as the ID Token's lifetime. (The authoritative value for the ID Token's lifetime is the `exp` claim inside the token itself, per OIDC Core §2.)

The `IdTokenTokenResponse` subclass (`response_type=id_token token`) *does* return an access token, so `expires_in` is appropriate there.

### Expected behavior

`expires_in` should only be included in responses that carry an access token (`id_token token`), not in the pure `id_token` response.

55728:
[OpenID Connect Core 1.0 §3.2.2.5](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthResponse) defines the response parameters for the Implicit Flow. `expires_in` is defined as the expiration time of the **Access Token** in seconds.

When `response_type=id_token` (the pure ID Token response), no access token is returned — the existing spec (`id_token_response_spec.rb`) already verifies that `access_token` is excluded. However, `expires_in` is still included in the response body:

```ruby
# lib/doorkeeper/oauth/id_token_response.rb
def body
  {
    expires_in: auth.token.expires_in_seconds,  # ← no access_token in this response
    state: pre_auth.state,
    id_token: id_token.as_jws_token
  }
end
```

Including `expires_in` without an accompanying `access_token` is misleading — a client could misinterpret it as the ID Token's lifetime. (The authoritative value for the ID Token's lifetime is the `exp` claim inside the token itself, per OIDC Core §2.)

### Fix

Move `expires_in` from `IdTokenResponse#body` into `IdTokenTokenResponse#body`, where an access token **is** returned. The pure `id_token` response now carries only `state` and `id_token`.

### Backward compatibility

This is technically a **breaking change** to the `response_type=id_token` response shape: the `expires_in` parameter is no longer present.

However:
- `expires_in` in this response was always meaningless per the spec (it represents the Access Token lifetime, and no `access_token` is returned).
- The `response_type=id_token token` response is unaffected — `expires_in` is still returned there, where it is semantically correct.
- Clients that were (incorrectly) using this value as the ID Token lifetime should switch to the `exp` claim inside the ID Token itself, which is the authoritative value per OIDC Core §2.

### Tests

- Updated `id_token_response_spec.rb`: body and redirect_uri expectations no longer include `expires_in`, plus explicit negative assertions.
- Updated `id_token_token_response_spec.rb`: body and redirect_uri expectations now include `expires_in` via `super.merge(...)` in `IdTokenTokenResponse#body`.

Closes #253

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - [#246] Fix `at_hash` to use correct hash algorithm based on `signing_algorithm`
 * [#250] Return configured `issuer` instead of `root_url` in WebFinger response (thanks to @sato11 for the original work in #172)
 - [#248] Fix `max_age` always triggering reauthentication when `auth_time_from_resource_owner` returns Integer
+- [#254] **Breaking:** Omit `expires_in` from the `response_type=id_token` response (OIDC Core §3.2.2.5 — `expires_in` represents the Access Token lifetime; it is still returned for `response_type=id_token token`)
 
 ## v1.9.0 (2026-03-16)
 
--- a/lib/doorkeeper/oauth/id_token_response.rb
+++ b/lib/doorkeeper/oauth/id_token_response.rb
@@ -19,7 +19,6 @@ module Doorkeeper
 
       def body
         {
-          expires_in: auth.token.expires_in_seconds,
           state: pre_auth.state,
           id_token: id_token.as_jws_token
         }
--- a/lib/doorkeeper/oauth/id_token_token_response.rb
+++ b/lib/doorkeeper/oauth/id_token_token_response.rb
@@ -6,7 +6,8 @@ module Doorkeeper
       def body
         super.merge({
           access_token: auth.token.token,
-          token_type: auth.token.token_type
+          token_type: auth.token.token_type,
+          expires_in: auth.token.expires_in_seconds
         })
       end
     end
--- a/spec/lib/oauth/id_token_response_spec.rb
+++ b/spec/lib/oauth/id_token_response_spec.rb
@@ -37,24 +37,30 @@ describe Doorkeeper::OAuth::IdTokenResponse do
   let(:id_token) { Doorkeeper::OpenidConnect::IdToken.new(token, pre_auth) }
 
   describe '#body' do
-    it 'return body response for id_token' do
+    it 'returns the id_token and state only (no expires_in per OIDC Core §3.2.2.5)' do
       expect(subject.body).to eq({
-        expires_in: auth.token.expires_in_seconds,
         state: pre_auth.state,
         id_token: id_token.as_jws_token
       })
     end
+
+    it 'does not include expires_in' do
+      expect(subject.body).not_to have_key(:expires_in)
+    end
   end
 
   describe '#redirect_uri' do
-    it 'includes expires_in, id_token and state' do
-      expect(subject.redirect_uri).to include("#{pre_auth.redirect_uri}#expires_in=#{auth.token.expires_in_seconds}&" \
-        "state=#{pre_auth.state}&" \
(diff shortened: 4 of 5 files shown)


nbulaj (commented):
LGTM as well ✅ 

I think we need to create/edit a page about upgrade of the gem with the notes like it's done for a core gem: https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions

And I see you already added a note in the changelog, that's 👍

55728:
Thanks! I drafted a Wiki page based on Doorkeeper’s Migration from old versions format:

<details>

<summary>Draft: Migration from old versions</summary>

# Migration from old versions

The intent of this document is to make migration of breaking changes as easy as possible. This page follows the same convention as the [core Doorkeeper gem](https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions).

Please also check the [CHANGELOG.md](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/CHANGELOG.md) for a full list of changes before finalizing the upgrade process.

## From 1.9.x to 1.10.x

### Breaking changes:

- **`expires_in` removed from pure `id_token` response** ([#254](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/254))
  
  When using the Implicit Flow with `response_type=id_token`, the response body and redirect URI no longer include the `expires_in` parameter.
  
  Per [OpenID Connect Core 1.0 §3.2.2.5](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthResponse), `expires_in` represents the Access Token lifetime — but no access token is issued in a pure `id_token` response, so this value was always semantically incorrect.
  
  **Who is affected?**
  
  Only clients using `response_type=id_token`. The `response_type=id_token token` response is **not** affected — `expires_in` is still returned there alongside the access token.
  
  **What to do:**
  
  If your client was reading `expires_in` from a pure `id_token` response (e.g. to determine the ID Token’s lifetime), switch to the `exp` claim inside the ID Token itself, which is the authoritative expiration value per [OIDC Core §2](https://openid.net/specs/openid-connect-core-1_0.html#IDToken).

</details>

I can create the Wiki page directly if I have write access, or feel free to use/adjust this as you see fit. The version number is tentative — happy to update it to match the actual release.

nbulaj:
This looks correct to me @55728 ,thanks! I created [a page](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/wiki/Migration%E2%80%90from%E2%80%90old%E2%80%90versions) . I believe you should have an access to edit it :eyes: Please feel free to edit it. In the meantime I'm gonna merge this one, thanks! I will update versions later when releasing.

image

nbulaj (approved):

55728:
Thanks for creating the page and merging! I’ve added the migration notes for #254 👀✨

nbulaj:
No-no. Thank you!

PR #256, opened 2026-04-24
Accept non-callable values for the `protocol` config option

### LINKED ISSUE #255: `protocol` config raises `NoMethodError` when set to a symbol or string
Other configuration options (`issuer`, `signing_algorithm`, `signing_key`, `expiration`) accept either a plain value or a callable (block/lambda), resolved via a `respond_to?(:call)` check. The `protocol` option is inconsistent — `DiscoveryController#protocol` calls `.call` unconditionally:

```ruby
# app/controllers/doorkeeper/openid_connect/discovery_controller.rb
def protocol
  Doorkeeper::OpenidConnect.configuration.protocol.call
end
```

Configuring `protocol :https` or `protocol 'https'` produces a `NoMethodError` at request time:

```
NoMethodError: undefined method 'call' for an instance of Symbol
```

The README does not explicitly state that `protocol` requires a block, and the inconsistency with other options is surprising.

### Expected behavior

`protocol` should accept both plain values and callables, matching the pattern used by other configuration options:

```ruby
def protocol
  configured = Doorkeeper::OpenidConnect.configuration.protocol
  configured.respond_to?(:call) ? configured.call : configured
end
```

55728:
Other configuration options (`issuer`, `signing_algorithm`, `signing_key`, `expiration`) accept either a plain value or a callable (block/lambda) and resolve the two shapes with a `respond_to?(:call)` check. For example, `issuer` can be configured as:

```ruby
issuer 'https://example.com'
# or
issuer { |resource_owner, application| "https://#{application.name}.example.com" }
```

The `protocol` option is inconsistent — `DiscoveryController#protocol` calls `.call` unconditionally:

```ruby
def protocol
  Doorkeeper::OpenidConnect.configuration.protocol.call
end
```

Configuring `protocol :https` (a symbol) or `protocol 'https'` (a string) produces a `NoMethodError` at request time.

### Fix

Guard with `respond_to?(:call)`, matching the pattern used by other options:

```ruby
def protocol
  configured = Doorkeeper::OpenidConnect.configuration.protocol
  configured.respond_to?(:call) ? configured.call : configured
end
```

### Tests

Added specs for both symbol and string values to `discovery_controller_spec.rb`.

Closes #255

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [#248] Fix `max_age` always triggering reauthentication when `auth_time_from_resource_owner` returns Integer
 - [#254] **Breaking:** Omit `expires_in` from the `response_type=id_token` response (OIDC Core §3.2.2.5 — `expires_in` represents the Access Token lifetime; it is still returned for `response_type=id_token token`)
 - [#252] Treat `auth_time_from_resource_owner` as optional in `IdToken` — omit `auth_time` claim when unconfigured instead of raising `InvalidConfiguration`
+- [#256] Accept non-callable values (symbol / string) for the `protocol` config option, matching the pattern used by `issuer` / `signing_algorithm` / `signing_key` / `expiration`
 
 ## v1.9.0 (2026-03-16)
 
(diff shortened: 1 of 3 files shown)


nbulaj (approved):
Makes sense 👍

nbulaj:
Merging, thanks!

PR #260, opened 2026-04-26
Skip IdToken construction on authorization code grants without `openid` scope

### LINKED ISSUE #259: `AuthorizationCodeRequest` constructs IdToken on every authorization code grant, even without `openid` scope
Same issue as #257, but in the authorization code flow.

`AuthorizationCodeRequest#after_successful_response` unconditionally instantiates `Doorkeeper::OpenidConnect::IdToken`:

```ruby
# lib/doorkeeper/openid_connect/oauth/authorization_code_request.rb
def after_successful_response
  super

  nonce =
    if (openid_request = grant.openid_request)
      openid_request.destroy!
      openid_request.nonce
    end

  id_token = Doorkeeper::OpenidConnect::IdToken.new(access_token, nonce)
  @response.id_token = id_token
end
```

`IdToken#initialize` eagerly invokes `resource_owner_from_access_token`, so non-OIDC authorization code exchanges (without the `openid` scope) pay for the resource owner lookup on every request. If `resource_owner_from_access_token` is not configured, the request fails outright.

As with the password grant (#257), `TokenResponse#body` already checks `token.includes_scope?('openid')` before serializing the `id_token` — so the constructed object is discarded anyway for non-OIDC grants.

### Expected behavior

The `IdToken` should only be constructed when the access token includes the `openid` scope, matching the guard already present in `TokenResponse#body`.

See also: #258 (the password grant fix, which [noted this follow-up](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/258#issue-4328313640)).

55728:
Follow-up to #258, as [noted there](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/258#issue-4328313640).

`AuthorizationCodeRequest#after_successful_response` unconditionally instantiates `Doorkeeper::OpenidConnect::IdToken` for every authorization code exchange:

```ruby
def after_successful_response
  super

  nonce =
    if (openid_request = grant.openid_request)
      openid_request.destroy!
      openid_request.nonce
    end

  id_token = Doorkeeper::OpenidConnect::IdToken.new(access_token, nonce)
  @response.id_token = id_token
end
```

`IdToken#initialize` eagerly invokes `resource_owner_from_access_token`, so non-OIDC authorization code exchanges (without the `openid` scope) pay for the resource owner lookup on every request. If `resource_owner_from_access_token` is not configured at all, the request fails outright.

Note that `TokenResponse#body` already checks `token.includes_scope?('openid')` before serializing the `id_token` — but the object is still constructed regardless.

### Fix

Guard the `IdToken` construction with the same scope check used in `TokenResponse#body` (identical approach to #258):

```ruby
def after_successful_response
  super

  if access_token.includes_scope?('openid')
    nonce =
      if (openid_request = grant.openid_request)
        openid_request.destroy!
        openid_request.nonce
      end

    id_token = Doorkeeper::OpenidConnect::IdToken.new(access_token, nonce)
    @response.id_token = id_token
  end
end
```

### Tests

* Existing test updated to explicitly set `scopes: 'openid'` on the access token. (The previous test passed without it because `IdToken.new` was invoked regardless of scope; with the guard, the fixture must explicitly grant `openid` to exercise the same code path.)
* New context for non-openid scope verifying `IdToken` is not instantiated and `response.id_token` is `nil`.

Closes #259

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - [#252] Treat `auth_time_from_resource_owner` as optional in `IdToken` — omit `auth_time` claim when unconfigured instead of raising `InvalidConfiguration`
 - [#256] Accept non-callable values (symbol / string) for the `protocol` config option, matching the pattern used by `issuer` / `signing_algorithm` / `signing_key` / `expiration`
 - [#258] Skip `IdToken` construction on password grants without the `openid` scope
+- [#259] Skip `IdToken` construction on authorization code grants without the `openid` scope
 
 ## v1.9.0 (2026-03-16)
 
--- a/lib/doorkeeper/openid_connect/oauth/authorization_code_request.rb
+++ b/lib/doorkeeper/openid_connect/oauth/authorization_code_request.rb
@@ -9,11 +9,11 @@ module Doorkeeper
         def after_successful_response
           super
 
-          nonce =
-            if (openid_request = grant.openid_request)
-              openid_request.destroy!
-              openid_request.nonce
-            end
+          openid_request = grant.openid_request
+          nonce = openid_request&.nonce
+          openid_request&.destroy!
+
+          return unless access_token.includes_scope?('openid')
 
           id_token = Doorkeeper::OpenidConnect::IdToken.new(access_token, nonce)
           @response.id_token = id_token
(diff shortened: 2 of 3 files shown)


nbulaj (approved):
Also makes sense, thanks

55728:
Thanks for the approval! 👀✨

I've addressed both Copilot suggestions:

- `openid_request` cleanup (`destroy!`) is now always performed when the record exists, regardless of scope — this prevents orphaned `oauth_openid_requests` rows.
- Added a test case where the grant has an `openid_request` but the token lacks `openid` scope, verifying both cleanup and that `IdToken` is not constructed.

Force-pushed to the branch. Let me know if you'd like me to adjust anything!

nbulaj (approved):
We're good to go now 👍

PR #261, opened 2026-04-26
Fix obsolete RuboCop RSpec configuration

55728:
`bundle exec rubocop` currently fails due to outdated settings in `.rubocop.yml`:

1. `require:` → `plugins:`
   (rubocop-rspec now uses the plugin system)

2. `RSpec/FilePath` → `RSpec/SpecFilePathFormat` + `RSpec/SpecFilePathSuffix`
   (the cop was split and the old name is no longer recognized)

3. Remove `Capybara/FeatureMethods`
   (moved to rubocop-capybara, which is not in the Gemfile)

After this change, `bundle exec rubocop` runs without configuration errors.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,5 @@
 inherit_from: .rubocop_todo.yml
-require:
+plugins:
   - rubocop-performance
   - rubocop-rails
   - rubocop-rspec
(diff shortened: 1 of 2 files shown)


nbulaj (approved):
I think nobody updated rubocop and the codebase for a while for this project so we can expect a lot of violations  😶‍🌫️

55728:
Yeah, running `bundle exec rubocop` confirms **272 offenses (145 autocorrectable)** 😅

I'm happy to follow up with additional PRs to tackle these — maybe something like:

1. **Auto-correctable offenses** (`rubocop -a`) as a quick win
2. **`NewCops: enable`** to settle the pending cop warnings
3. **Remaining manual fixes** (Documentation cops, Metrics thresholds, etc.) in smaller batches

Let me know if that sounds good, or if you'd prefer a different approach!

nbulaj:
Yes, sounds like a plan :eyes: :+1:

Issue #249, opened 2026-04-23
Dynamic client registration: several RFC 7591 compliance gaps


55728:
### Description

While reviewing the dynamic client registration implementation, I found several areas where the behavior diverges from [RFC 7591](https://datatracker.ietf.org/doc/html/rfc7591). I'm raising them together since they all affect the same endpoint and the fixes may overlap.

#### 1. All clients are registered as public (`confidential: false`)

The controller hardcodes `confidential: false`:

```ruby
# app/controllers/doorkeeper/openid_connect/dynamic_client_registration_controller.rb
def application_params
  {
    name: params.dig(:client_name),
    redirect_uri: params.dig(:redirect_uris) || [],
    scopes: params.dig(:scope),
    confidential: false,  # ← always public
  }
end
```

Per [RFC 7591 §2](https://datatracker.ietf.org/doc/html/rfc7591#section-2), the default `token_endpoint_auth_method` is `client_secret_basic`, which implies the client is **confidential**. The current behavior creates a mismatch:

1. The discovery response advertises `token_endpoint_auth_methods_supported: ["client_secret_basic", "client_secret_post"]`
2. The registration response includes a `client_secret`
3. But the client is registered as **public**, so Doorkeeper does not require `client_secret` at the token endpoint

This means anyone with a `client_id` can exchange tokens without the secret, effectively bypassing client authentication.

A possible fix: determine `confidential` based on the requested `token_endpoint_auth_method` (`none` → public, otherwise → confidential). I'd like to understand whether the current `confidential: false` was an intentional design choice before submitting a PR.

#### 2. Client-requested parameters are silently ignored

The registration endpoint reads only `client_name`, `redirect_uris`, and `scope`. Several parameters defined in [RFC 7591 §2](https://datatracker.ietf.org/doc/html/rfc7591#section-2) are silently ignored:

- **`token_endpoint_auth_method`** — always responds with a hardcoded list
- **`application_type`** — always responds with `"web"`
- **`response_types`** / **`grant_types`** — always returns the server's global configuration
- **`client_uri`**, **`logo_uri`**, **`tos_uri`**, **`policy_uri`**, **`contacts`** — client metadata is discarded

[...]

PR #262, opened 2026-04-29
Determine client confidentiality from `token_endpoint_auth_method` in dynamic registration

55728:
Refs #249 (point 1 — `confidential` hardcoded to `false`).

## Problem

The dynamic client registration endpoint hardcodes `confidential: false` for all registered clients. Per [RFC 7591 §2](https://datatracker.ietf.org/doc/html/rfc7591#section-2), the default `token_endpoint_auth_method` is `client_secret_basic`, which implies a **confidential** client. The current behavior creates a mismatch: the discovery response advertises `client_secret_basic`/`client_secret_post`, a `client_secret` is issued, but Doorkeeper does not actually require it at the token endpoint.

## Changes

**Controller:**

- Derive `confidential` from the client-requested `token_endpoint_auth_method` (`"none"` → public, otherwise → confidential). Defaults to `client_secret_basic` when omitted, matching RFC 7591 §2.
- Return `token_endpoint_auth_method` (singular) in the registration response instead of `token_endpoint_auth_methods_supported` (plural), aligning with [RFC 7591 §3.2.1](https://datatracker.ietf.org/doc/html/rfc7591#section-3.2.1) which defines the per-client field as singular.
- Omit `client_secret` from the response when the client is public (`token_endpoint_auth_method: "none"`).
- Reject unsupported auth methods (e.g. `private_key_jwt`) with `invalid_client_metadata` / 400, per [RFC 7591 §3.2.2](https://datatracker.ietf.org/doc/html/rfc7591#section-3.2.2).

**Tests:** expanded from 2 to 6 examples covering each supported auth method, the public client path, unsupported method rejection, and the existing redirect URI validation.

## ⚠️ Breaking changes

1. **Default confidentiality flipped:** Previously all dynamically registered clients were **public** (`confidential: false`). After this change, the default is **confidential** (`confidential: true`), matching RFC 7591. Clients that require public registration must now explicitly send `token_endpoint_auth_method: "none"`.

2. **Response field renamed:** The registration response previously returned `token_endpoint_auth_methods_supported` (plural). This is a discovery metadata field name ([OIDC Discovery §3](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata)); the per-client registration response should use `token_endpoint_auth_method` (singular) per [RFC 7591 §3.2.1](https://datatracker.ietf.org/doc/html/rfc7591#section-3.2.1). Clients parsing the old field name will need to update.

## Note

`SUPPORTED_AUTH_METHODS` is currently a static constant. Ideally this should be derived from `Doorkeeper.configuration.client_credentials_methods` to stay consistent with the discovery endpoint ([`discovery_controller.rb:82–85`](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/app/controllers/doorkeeper/openid_connect/discovery_controller.rb#L82-L85)). I kept it simple for now and can align them in a follow-up if preferred.

## Out of scope

Points 2 (ignored client metadata parameters) and 3 (unauthenticated endpoint / Initial Access Token) from #249 are intentionally left for separate PRs to keep this change focused.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - [#258] Skip `IdToken` construction on password grants without the `openid` scope
 - [#259] Skip `IdToken` construction on authorization code grants without the `openid` scope
 - [#261] Fix obsolete RuboCop configuration (`require:` → `plugins:`, `RSpec/FilePath` split, remove `Capybara/FeatureMethods`)
+- [#262] **Breaking:** Determine client confidentiality from `token_endpoint_auth_method` in dynamic registration — previously all clients were registered as public (`confidential: false`); now defaults to confidential (`client_secret_basic`) per RFC 7591 §2. Also renames response field from `token_endpoint_auth_methods_supported` to `token_endpoint_auth_method` per RFC 7591 §3.2.1 (fixes #249). Derive supported auth methods from `Doorkeeper.configuration.client_credentials_methods` via shared `TokenEndpointAuthMethodsMixin`, keeping discovery and registration in sync.
 
 ## v1.9.0 (2026-03-16)
 
--- a/app/controllers/concerns/doorkeeper/openid_connect/token_endpoint_auth_methods_mixin.rb
+++ b/app/controllers/concerns/doorkeeper/openid_connect/token_endpoint_auth_methods_mixin.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module Doorkeeper
+  module OpenidConnect
+    module TokenEndpointAuthMethodsMixin
+      private
+
+      # Maps Doorkeeper's client_credentials configuration to OIDC auth method names.
+      # Shared by DiscoveryController and DynamicClientRegistrationController
+      # to keep advertised and accepted auth methods in sync.
+      AUTH_METHOD_MAPPING = {
+        from_basic: 'client_secret_basic',
+        from_params: 'client_secret_post',
+      }.freeze
+
+      def token_endpoint_auth_methods_supported(doorkeeper = ::Doorkeeper.configuration)
+        doorkeeper.client_credentials_methods.filter_map { |method| AUTH_METHOD_MAPPING[method] }
+      end
+    end
+  end
+end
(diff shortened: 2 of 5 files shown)


55728:
**Addressed Copilot's review feedback (d36682c)**

@nbulaj Per your 👍 on Copilot's comment, I replaced the hardcoded `SUPPORTED_AUTH_METHODS` constant with a shared `TokenEndpointAuthMethodsMixin` that derives supported methods from `Doorkeeper.configuration.client_credentials_methods`.

**What changed:**

- **New:** `app/controllers/concerns/doorkeeper/openid_connect/token_endpoint_auth_methods_mixin.rb` — holds the `from_basic` / `from_params` → OIDC string mapping (same mapping that `DiscoveryController` was using inline).
- **`DiscoveryController`** — includes the mixin; the inline `token_endpoint_auth_methods_supported` private method is removed.
- **`DynamicClientRegistrationController`** — includes the mixin; `SUPPORTED_AUTH_METHODS` constant removed. Validation and error messages now use the dynamic list. `"none"` is appended on the registration side only, since it's valid for public client registration per RFC 7591 but isn't a token endpoint auth method the server advertises in discovery.
- **Specs** — added a `"when Doorkeeper client_credentials is restricted to :from_basic"` context: `client_secret_basic` accepted, `client_secret_post` rejected with `invalid_client_metadata`, `"none"` still accepted for public clients.

This guarantees that discovery and registration always agree on which auth methods are supported, regardless of how `client_credentials` is configured.

221 examples, 0 failures.

nbulaj:
hey @55728 , isn't it the same as in https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/263 ? :eyes:

55728:
Yes, they address the same root cause from different angles — #262 started as an RFC 7591 compliance fix for #249 point 1, and #263 came from our email discussion framing it as a security issue (by_uid_and_secret bypass). I should have closed #262 when I opened #263. Closing this one in favor of #263.

55728:
Sorry for the confusion and the extra noise, @nbulaj — I should have connected the dots between #249 and the email thread earlier. Won't happen again.

PR #268, opened 2026-05-06
Update Dynamic Client Registration README for validated metadata parameters

55728:
Follow-up to #266 — update the Dynamic Client Registration section to reflect that `application_type`, `response_types`, and `grant_types` are now validated and echoed back in the registration response, rather than silently ignored.

### Changes

- Add `application_type`, `response_types`, and `grant_types` to the supported parameters table
- Remove the paragraph stating these parameters return server-side defaults regardless of client input (no longer accurate post-#266)
- Update the #249 reference to note that only `client_uri`, `logo_uri`, `contacts`, etc. remain unsupported

Refs #249.


nbulaj (approved):
LGTM, thanks!

Issue #238, opened 2026-03-20
issuer block params not consistent


szechyjs:
The parameters that are passed to `Doorkeeper::OpenidConnect.configuration.issuer` block are not consistent.

In [DiscoveryController](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/225e3194da02aeb362d22c328b2e4babe66b32aa/app/controllers/doorkeeper/openid_connect/discovery_controller.rb#L134) it passes in the current `request` as the only parameter. Then in [IdToken](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/225e3194da02aeb362d22c328b2e4babe66b32aa/lib/doorkeeper/openid_connect/id_token.rb#L46C63-L46C105) it passes in `resource_owner, application`.

Ideally these should both pass in the same parameters, that way the config can stay consistent and not have to check parameter types.

```rb
# default from generator
issuer do |resource_owner, application|
  'issuer string'
end

# what works for discovery controller
issuer do |request|
  request.base_url
end

# proposed
issuer do |resource_owner, application, request|
  request.base_url
end
```

55728:
Hi, I’d like to take this on!

I’ll unify the issuer block parameters by passing resource_owner, application, and request from both DiscoveryController and IdToken, with an arity check for backward compatibility.

PR #228, opened 2025-11-04
chore(README.md): improve docs

stakach:
I recently integrated this into an application and found this worked for us. The existing documentation was a little confusing

PR #270, opened 2026-05-08
Document the unified issuer block signature in README

55728:
## Summary

Update the README description of the `issuer` configuration option to reflect the unified block signature introduced in #238 / 2225ce8e (`Doorkeeper::OpenidConnect.resolve_issuer`).

Before that change, the `issuer` block was invoked with different arguments depending on the caller — `(resource_owner, application)` from `IdToken` and `(request)` from `DiscoveryController` — and the README only described that split behavior, pointing to a now-stale line in `discovery_controller.rb`. After the change, a single block can declare `do |resource_owner, application, request|` and work correctly in both contexts, but the README never advertised this form.

This PR:

- Adds a runnable example of the recommended 3-argument form, mirroring the generator template at `lib/generators/doorkeeper/openid_connect/templates/initializer.rb`.
- Explains why a single block can serve both ID token issuance and the discovery endpoint.
- Documents the legacy arity-0/1/2 forms that remain supported for backward compatibility.
- Removes the broken source link (`discovery_controller.rb#L123` no longer points at the `issuer` method).

## Related

- Implements documentation for the change made in #238 (commit 2225ce8e).
- Supersedes #228, which attempted to document the pre-unification behavior with an arity-2 example. The author has not responded to maintainer feedback since, and the example there relied on the pre-unification dispatch quirk (DiscoveryController calling with a single positional arg, leaving `application` nil) — after #238 / 2225ce8e, an arity-2 block always receives `(resource_owner, application)` regardless of context, so that example would `NoMethodError` on the discovery endpoint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - [#267] Add `authorize_dynamic_client_registration` config option to gate the dynamic client registration endpoint per RFC 7591 §3.1 — when set to a callable, the block is evaluated in the controller scope (with access to `request`, `params`, `request.headers`, etc.) and falsy return values reject the request with `401 invalid_token`. Default is `nil` so the endpoint remains open for backward compatibility; consumers should configure this to validate an Initial Access Token (or any other authorization scheme) before allowing client registration
 - [#268] Update Dynamic Client Registration README for validated metadata parameters
 - [#269] Document `authorize_dynamic_client_registration` in README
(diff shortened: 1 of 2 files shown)


nbulaj:
Thanks!

55728:
Thank you for the review and merge, @nbulaj! 🎉

Now that this is merged, I believe #228 can be closed as well — this PR supersedes it. Happy to leave that call to you! 😊

PR #271, opened 2026-05-12
Add `auth_time_from_session` for per-session `max_age` enforcement

### LINKED ISSUE #150: `auth_time_from_session` instead of `auth_time_from_resource_owner`
To implement `auth_time` correctly, I propose `auth_time_from_sesion`

**Details**

The `auth_time_from_resource_owner` is designed to get the authentication time from the object of the resource_owner.
In many cases, developers are likely to implement getting the last login time of the resource owner from the Database.
This is also implied by the sample implementation of the template.
https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/bd684594e97229c10a7e2670508e7f7c27658604/lib/generators/doorkeeper/openid_connect/templates/initializer.rb#L21-L24

The authentication time stored in the database will be overwritten when the user sign in with another session.
Therefore, this leads to an insecure specification.
For example, if a user sign in 12 hours ago on PC browser and 1 hour ago on smartphone browser, and the user do the OIDC flow on PC, `auth_time` will be the time 1 hour ago. However `auth_time` should be the time the user logged in on your PC.
Even if RelyingParty requires that you authenticate within 6 hours using `max_age`, it will not re-authenticate.

**Proposal**

To solve this, my idea is as follows.

Add `auth_time_from_session` as a config instead of `auth_time_from_resource_owner`.

The [`handle_oidc_max_age_param`](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/bd684594e97229c10a7e2670508e7f7c27658604/lib/doorkeeper/openid_connect/helpers/controller.rb#L88)  will also be able to correctly determine for each session if it should be re-authenticated.
And, when creating an `openid_request`, get the auth_time from the session and store the time in the DB with the nonce. This way, the authorization code and auth_time will be bound.Then, it will be possible to return the auth_time associated with the specified authorization code in the token request.

55728:
### Problem

`auth_time_from_resource_owner` derives `auth_time` from the resource owner object — in practice, the user's most recent login stored in the database. When a user holds multiple concurrent sessions (e.g. PC + smartphone), every session shares the same `auth_time`. This means a stale session can satisfy an RP's `max_age` requirement by piggy-backing on a fresh login from another device, which is a violation of OIDC Core §3.1.2.1.

See #150 for the original report.

### Solution

Add a new `auth_time_from_session` config option that receives `(session, request)` in the controller scope, allowing deployments to return the auth time of the *current* session.

The core change is two points:

1. **`config.rb`**: Add `option :auth_time_from_session, default: nil`
2. **`controller.rb`**: Extract `resolve_oidc_auth_time` — prefers `auth_time_from_session` when configured, falls back to `auth_time_from_resource_owner` with a one-time deprecation warning

### Backward compatibility

Existing users who only configure `auth_time_from_resource_owner` are **not affected** — the only change they will see is a one-time deprecation warning on stderr (once per process).

### What's NOT in scope

- The `auth_time` claim on the ID Token still uses `auth_time_from_resource_owner`, because `IdToken` is constructed from an access token and has no access to the session/request.
- README update is intentionally deferred to a follow-up PR to keep this diff focused.

Closes #150

### DIFF
--- a/lib/doorkeeper/openid_connect/helpers/controller.rb
+++ b/lib/doorkeeper/openid_connect/helpers/controller.rb
@@ -4,6 +4,29 @@ module Doorkeeper
   module OpenidConnect
     module Helpers
       module Controller
+        # Emit `auth_time_from_resource_owner` deprecation at most once per process
+        # to avoid spamming logs on every authorize request with `max_age`.
+        @auth_time_from_resource_owner_deprecation_warned = false
+
+        def self.warn_auth_time_from_resource_owner_deprecation
+          return if @auth_time_from_resource_owner_deprecation_warned
+
+          @auth_time_from_resource_owner_deprecation_warned = true
+          warn "DEPRECATION WARNING: `auth_time_from_resource_owner` is deprecated for " \
+               "`max_age` enforcement because it cannot distinguish between concurrent " \
+               "sessions of the same user, which is a security issue (see " \
+               "https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/150). " \
+               "Please configure `auth_time_from_session` to derive auth_time from the " \
+               "current session instead. The `auth_time_from_resource_owner` callback " \
+               "continues to be used as a fallback and for the `auth_time` claim on the " \
+               "ID Token."
+        end
+
+        # Reset the deprecation flag (test helper).
+        def self.reset_auth_time_deprecation_warning!
+          @auth_time_from_resource_owner_deprecation_warned = false
+        end
+
         private
 
         # FIXME: remove after Doorkeeper will merge it
@@ -99,10 +122,7 @@ module Doorkeeper
           max_age = params[:max_age].to_i
           return unless (params[:max_age].to_s == "0" || max_age > 0) && owner
 
-          auth_time = instance_exec(
-            owner,
-            &Doorkeeper::OpenidConnect.configuration.auth_time_from_resource_owner
-          )
+          auth_time = resolve_oidc_auth_time(owner)
 
           # Normalize non-Time values (e.g. an Integer epoch) so that the
           # subtraction below yields a Float of elapsed seconds rather than a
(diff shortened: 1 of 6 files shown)


lib/doorkeeper/openid_connect/helpers/controller.rb
```diff
@@ -4,6 +4,29 @@ module Doorkeeper
...
       module Controller
+        # Emit `auth_time_from_resource_owner` deprecation at most once per process
+        # to avoid spamming logs on every authorize request with `max_age`.
+        @auth_time_from_resource_owner_deprecation_warned = false
+
+        def self.warn_auth_time_from_resource_owner_deprecation
+          return if @auth_time_from_resource_owner_deprecation_warned
+
+          @auth_time_from_resource_owner_deprecation_warned = true
+          warn "DEPRECATION WARNING: `auth_time_from_resource_owner` is deprecated for " \
```
nbulaj, line 15: We can add a `deprecated: { message: "..." }` for the `auth_time_from_resource_owner` , won't it work? 👀

55728: Good idea! 😊 Doorkeeper's `option` DSL `deprecated:` would be much cleaner than the hand-rolled warning.

One thing I'd like to clarify: the timing is different between the two approaches.

- **`deprecated:` on the option**: warns at **boot time** when the user configures `auth_time_from_resource_owner` in the initializer
- **Current implementation**: warns at **runtime** when `max_age` is actually evaluated and `auth_time_from_session` is not configured

Since `auth_time_from_resource_owner` is still needed for the `auth_time` claim on the ID Token (where session/request aren't available), deprecating the option itself might be too broad — users who configure both `auth_time_from_session` (for `max_age`) and `auth_time_from_resource_owner` (for the claim) would still see the warning at boot.

What do you think is the better approach here? Should we:

1. Use `deprecated:` and accept the boot-time warning, or
2. Keep the runtime warning scoped to the `max_age` fallback path?

Happy to adjust either way 👀

nbulaj: Hm good question :eyes: So there is still a case when configuring `auth_time_from_resource_owner` is  a good option, right?

> still needed for the auth_time claim on the ID Token (where session/request aren't available)

In such case we shouldn't deprecate the option itself, yes :+1:

55728: Makes sense! Since `auth_time_from_resource_owner` is still the only source for the ID Token `auth_time` claim, deprecating the option itself would be too broad.

I'll keep the current runtime warning (only fires on the max_age fallback path) and rebase to resolve the CHANGELOG conflict. Thanks for the thoughtful review 😊

nbulaj (approved):
Thanks! LGTM

PR #276, opened 2026-05-17
Get RuboCop to zero offenses (176 → 0)

55728:
## Summary

Follow-up to #264 — resolves the remaining offenses to reach zero.

Bring RuboCop offenses from **176 → 0** with no test regressions (232 examples, 0 failures — verified on two random seeds from a clean DB).

The second commit adds a dedicated RuboCop job to CI so that the zero-offense baseline is enforced on every push and pull request.

## Motivation

`.rubocop_todo.yml` has not been updated since 2020. Running `rubocop` on the current codebase reports 176 offenses, which makes it easy to miss new violations in PRs. This patch brings the count to zero so CI can enforce a clean baseline going forward.

### Real code fixes

| Cop | Fix |
|-----|-----|
| **Lint/MissingSuper** | Call `super()` in `IdTokenResponse#initialize` |
| **Rails/Output** | Replace `puts` → `warn` for deprecation notices (stderr, not stdout) |
| **Naming/HeredocDelimiterNaming** | `EOL` → `RSA_PRIVATE_KEY` / `EC_PRIVATE_KEY` |
| **RSpec/DescribedClass** | `Doorkeeper::OpenidConnect.configure` → `described_class.configure` |
| **RSpec/ExpectChange** | `change { Klass.count }` → `change(Klass, :count)` |
| **RSpec/ReceiveMessages** | Collapse chained `receive` stubs into `receive_messages` |
| **Style/RedundantInterpolation** | `"#{Rails.root.join(...)}"` → `.to_s` |
| **Style/FrozenStringLiteralComment** | Add missing magic comment |

### Safe autocorrects applied

All autocorrects from `rubocop -A` were reviewed; safe cops were applied as-is, and vetted unsafe cops (`RSpec/DescribedClass`, `RSpec/ExpectChange`, `RSpec/ReceiveMessages`) were verified manually.

### Behaviour-breaking autocorrects avoided

Two autocorrects were **reverted** after causing test failures:

1. **RSpec/LeadingSubject** — mangled the version-conditional `subject` in `password_access_token_request_spec.rb` (dropped the `else` branch). Restored the original and added an inline `# rubocop:disable RSpec/LeadingSubject`.

2. **Style/SymbolProc** — rewrote the Doorkeeper config DSL blocks (`subject`, `auth_time_from_resource_owner`, `claim`) to `&:sym`. These DSL methods pass extra arguments, so the strict-arity lambda raises `ArgumentError`. Restored explicit blocks and excluded the files in `.rubocop_todo.yml`.

### Configuration updates (`.rubocop.yml` / `.rubocop_todo.yml`)

- **Disabled:** `Style/Documentation`, `Naming/VariableNumber`, and noisy RSpec structural cops (`ContextWording`, `DescribeMethod`, `ExpectInHook`, `IndexedLet`, `SubjectStub`, `StubbedMock`)
- **Raised Max:** `Metrics/MethodLength` 15 → 33, `Metrics/AbcSize` → 35, `Metrics/CyclomaticComplexity` → 13, `Metrics/PerceivedComplexity` → 10, `RSpec/MultipleMemoizedHelpers` → 7
- **Excluded from Layout/LineLength:** legacy library files with long deprecation / I18n strings

All config changes follow the existing `.rubocop_todo.yml` convention (values derived from `rubocop --auto-gen-config`).

## Behavioural notes for existing users

This patch is mostly spec / config cleanup, but two fixes touch runtime
code:

1. **`puts` → `warn` for deprecation notices** (`config.rb`) — `jws_public_key` and `jws_private_key` deprecation messages now go to **stderr** instead of stdout. This is the standard Ruby / Rails convention for warnings. If you were capturing stdout to detect deprecations, you will need to capture stderr instead.

2. **`super()` added to `IdTokenResponse#initialize`** — `IdTokenResponse` inherits from `BaseResponse`, which does not define its own `initialize`, so `super()` resolves to `Object#initialize` (a no-op). This change silences the `Lint/MissingSuper` cop with **no behavioural effect**. Note that bare `super` (without parens) would forward the three arguments and raise `ArgumentError`; the explicit `super()` avoids this.

Neither change alters the gem's public interface or return values.

## Verification

```
$ bundle exec rubocop
...
91 files inspected, no offenses detected

$ bundle exec rspec --seed 12345
232 examples, 0 failures

$ bundle exec rspec --seed 55728
232 examples, 0 failures
```

>[!TIP]
>As a follow-up I'm planning to open a separate PR to upgrade `actions/checkout` from v3 → v6 and pin it to a commit SHA for supply-chain safety. `ruby/setup-ruby` will stay at `@v1` per their upstream recommendation (the `v1` tag is intentionally a moving target that tracks new Ruby releases).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,13 +20,13 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        os: [ ubuntu-latest ]
+        os: [ubuntu-latest]
         ruby:
-          - '3.1'
-          - '3.2'
-          - '3.3'
-          - '3.4'
-          - '4.0'
+          - "3.1"
+          - "3.2"
+          - "3.3"
+          - "3.4"
+          - "4.0"
           - head
         gemfile:
           - gemfiles/rails_7.0.gemfile
@@ -54,3 +54,25 @@ jobs:
       - name: Run tests
         timeout-minutes: 10
         run: bundle exec rake spec
+  lint:
+    name: RuboCop
+    runs-on: ubuntu-latest
+    env:
+      BUNDLE_PATH: /home/runner/bundle
+    if: |
+      !(   contains(github.event.pull_request.title,  '[ci skip]')
+        || contains(github.event.pull_request.title,  '[skip ci]'))
+    steps:
+      - name: Repo checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.4"
+
+      - name: Install dependencies
+        run: bundle install
+
+      - name: Run RuboCop
+        run: bundle exec rubocop
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -1,0 +1,37 @@
+name: rubocop
+on:
+  pull_request:
+    paths:
+      - 'lib/**/*.rb'
+      - 'app/**/*'
+      - 'spec/**/*.rb'
+      - 'bin/*'
+      - '.rubocop.yml'
+      - '.rubocop_todo.yml'
+      - 'Gemfile'
+      - 'Gemfile.lock'
+      - 'Rakefile'
+      - '*.gemspec'
+permissions:
+  contents: read
+  pull-requests: write
+jobs:
+  rubocop:
+    name: runner / rubocop
+    runs-on: ubuntu-latest
+    if: |
+      !(   contains(github.event.pull_request.title,  '[ci skip]')
+        || contains(github.event.pull_request.title,  '[skip ci]'))
+    env:
+      BUNDLE_ONLY: rubocop
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.4'
+          bundler-cache: true
+      - uses: reviewdog/action-rubocop@b6d5e953a5fc0bf3ab65254e77730ea2174d6d6d # v2.22.0
+        with:
+          reporter: github-pr-review # Default is github-pr-check
+          skip_install: true
+          use_bundler: true
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,6 +6,11 @@ plugins:
 AllCops:
   TargetRubyVersion: 3.1
   Exclude:
+    # CI installs gems into ./vendor/bundle (ruby/setup-ruby bundler-cache),
+    # which RuboCop would otherwise lint. Defining AllCops/Exclude overrides
+    # RuboCop's default vendor exclusion, so list it explicitly here too.
+    - "vendor/**/*"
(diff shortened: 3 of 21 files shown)


.github/workflows/ci.yml
```diff
@@ -53,3 +53,25 @@ jobs:
       - name: Run tests
         timeout-minutes: 10
         run: bundle exec rake spec
+  lint:
```
nbulaj, line 57: For doorkeeper we have an action which reports (comments) in the PR with the found violations, I think it will be good to have the one here as well

https://github.com/doorkeeper-gem/doorkeeper/blob/main/.github/workflows/rubocop.yml 

(I just don't have enough time to fine-tune it 🙈 )

55728: Thanks for the pointer @nbulaj! 😊 Added in f61f1ece076c0483a49896149b0ebd3050659ed3 + 9e88ad7dd48f90baada819656bf9b8e625df0803 — mirrors doorkeeper's rubocop.yml (reviewdog with github-pr-review, BUNDLE_ONLY: rubocop, the same paths filter, and the [ci skip]/[skip ci] guard).

A few notes:

- **Gemfile**: moved the rubocop gems into a group :development, :rubocop block (same as doorkeeper) so BUNDLE_ONLY: rubocop installs only what the job needs. Versions unchanged.

- **Action pinning**: reviewdog/action-rubocop is pinned to doorkeeper's vetted SHA (b6d5e953… = v2.22.0). actions/checkout stays at @v3 to match the existing ci.yml for now (the v3 → v6 + SHA pin is a natural follow-up PR); ruby/setup-ruby stays at @v1 per its upstream recommendation.

- The existing lint job in ci.yml is intentionally kept — it also runs on push, while this workflow runs only on pull_request.

While wiring this up I hit a real bug worth flagging: since .rubocop.yml defines its own AllCops/Exclude, it overrides RuboCop's built-in default that excludes `vendor/**/*`. ruby/setup-ruby with bundler-cache: true installs gems into ./vendor/bundle, so RuboCop linted the entire installed bundle (~2,800 files, 300k+ offenses) and the job failed. It didn't surface in ci.yml's lint job because that one sets BUNDLE_PATH=/home/runner/bundle (outside the workspace). Fixed in 9e88ad7dd48f90baada819656bf9b8e625df0803 by explicitly excluding `vendor/**/*` and `.bundle/**/*` — target file count and the existing zero-offense state are unchanged.

CI is green now. 🎉

nbulaj: Thanks!

nbulaj:
We can merge this one after rebase :+1:

nbulaj (approved):
LGTM ✅

55728:
Thank you for reviewing 😊 I rebased, and fixed a new offense!

nbulaj:
You can then just merge this one @55728 to check your permissions as well :rock:

55728:
Thank you @nbulaj 😊 — merging now to take my new permissions for a spin 🪨✨

Issue #207, opened 2024-06-09
Support dynamic signing_key


caherrerapa:
In some cases such as multi-tenant scenarios, we want to allow the signing key to be loaded based on the result of a block instead of a hardcoded string that would apply to the entire application instead of having different keys based on the context (e.g. different key per tenant)

nbulaj:
I can see `signing_key` supports a block as an argument for the config options, doesn't it solve your case @caherrerapa ?

https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/lib/doorkeeper/openid_connect.rb#L53-L58

55728:
To add to what @nbulaj mentioned — both `signing_key` and `signing_algorithm` accept callable objects, so they're evaluated on each call:

```ruby
signing_key -> { current_tenant.private_key }
signing_algorithm -> { current_tenant.algorithm }
```

@caherrerapa does this work for your use case? If so, we could close this one 😊

55728:
Closing as answered — both `signing_key` and `signing_algorithm` already accept callable objects, so per-tenant dynamic keys work out of the box. This is now documented in the README as well (#277). Feel free to reopen if this doesn't cover your use case. 😊

PR #277, opened 2026-05-17
Fix README inaccuracies and use constant-time comparison in DCR authorization example

55728:
Two documentation fixes in a single PR (one commit each):

**1. Fix README inaccuracies**

- `signing_algorithm`: describe it as the ID token **signing** algorithm (RS256 is signing, not encryption) and point to [ruby-jwt's algorithm docs](https://github.com/jwt/ruby-jwt#algorithms-and-usage) instead of the stale nov/json-jwt JWE wiki link.
- `discovery_url_options`: add `dynamic_client_registration` to the list of available endpoints (matches `DiscoveryController`).
- Routes: document the unconditionally-mounted `GET /.well-known/oauth-authorization-server` route.

**2. Use constant-time comparison in DCR authorization example**

The previous example compared the bearer token with `==`, which is vulnerable to a timing attack that could leak the Initial Access Token. Replace with `ActiveSupport::SecurityUtils.secure_compare` over SHA-256 digests so the comparison is both constant-time and fixed-length (avoiding token length leakage).

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,14 @@ The following settings are required in `config/initializers/doorkeeper_openid_co
 - `resource_owner_from_access_token`
   - Defines how to translate the Doorkeeper access token to a resource owner model.
 
+> [!Note]
+> Both `signing_key` and `signing_algorithm` also accept callable objects (e.g. a lambda), which are evaluated on each call — useful for multi-tenant setups where the key or algorithm varies per request:
+>
+> ```ruby
+> signing_key -> { current_tenant.private_key }
+> signing_algorithm -> { current_tenant.algorithm }
+> ```
+
 The following settings are optional, but recommended for better client compatibility:
 
 - `auth_time_from_resource_owner`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -143,7 +143,7 @@ The following settings are required in `config/initializers/doorkeeper_openid_co
   - You can generate a private key with the `openssl` command, see e.g. [Generate an RSA keypair using OpenSSL](https://en.wikibooks.org/wiki/Cryptography/Generate_a_keypair_using_OpenSSL).
   - You should not commit the key to your repository, but use an external file (in combination with `File.read`) and/or the [dotenv-rails](https://github.com/bkeepers/dotenv) gem (in combination with `ENV[...]`).
 - `signing_algorithm`
-  - The encryption type of the private key which defaults to `:rs256`. The list of supported algorithms can be found [here](https://github.com/nov/json-jwt/wiki/JWE#supported-algorithms)
+  - The signing algorithm used for the ID token, which defaults to `:rs256`. The list of supported algorithms can be found [here](https://github.com/jwt/ruby-jwt#algorithms-and-usage)
```
nbulaj, line 146: WDYT about adding a note that both options accept callable objects? Considering [this issue](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/207)

Then LGTM and also can be merged ✅

55728: Good idea — pushed a note with a multi-tenant example 👍

nbulaj (approved):

PR #280, opened 2026-05-18
Skip OIDC prompt/max_age handling when response is already committed

24c02:
## problem

  if a user hits an openid authorize request with `prompt=login` while not authenticated, the `reauthenticate_resource_owner` callback receives a non-model value and blows up. same issue applies to `prompt=none` and `max_age` -- anything  where `handle_oidc_prompt_param!` or `handle_oidc_max_age_param!` acts on the `owner` value.

  `authenticate_resource_owner!` uses `.tap` on the return value of `super`, but doesn't check whether a response was already committed. when the `resource_owner_authenticator` block redirects (the standard pattern for unauthenticated users), the `.tap` block still runs with whatever `redirect_to` returned — which is always truthy (`self.status` on rails 8, the response body string on rails 7). this passes the `if owner` guards in `handle_oidc_prompt_param!` and the configured callback tries to call model methods on a non-model value:

  `NoMethodError: undefined method 'sessions' for an instance of Integer`

  followed by a `DoubleRenderError` since the response was already committed by the authenticator's redirect.

  ## fix

  add a `performed?` guard at the top of the `.tap` block. if a redirect already happened, there's no authenticated owner to process OIDC params against.

  ## versions affected

  confirmed on doorkeeper-openid_connect 1.8.11 & 1.9.0, rails 8.0.4, doorkeeper 5.8.2. the underlying issue (no `performed?` check) affects all rails versions.

### DIFF
--- a/lib/doorkeeper/openid_connect/helpers/controller.rb
+++ b/lib/doorkeeper/openid_connect/helpers/controller.rb
@@ -15,6 +15,7 @@ module Doorkeeper
 
         def authenticate_resource_owner!
           super.tap do |owner|
+            next if performed?
             next unless oidc_authorization_request?
 
             handle_oidc_max_age_param!(owner)


lib/doorkeeper/openid_connect/helpers/controller.rb
```diff
@@ -15,6 +15,7 @@ def current_resource_owner
 
         def authenticate_resource_owner!
           super.tap do |owner|
+            next if performed?
             next unless oidc_authorization_request?
```
55728, line 19: Thanks for catching this — I reproduced the crash locally and it's a real bug. With the common host-app pattern `resource_owner_authenticator do current_user || redirect_to(login_url) end`, `super` returns the truthy value of `redirect_to` (status `Integer` on Rails 8, body `String` on Rails 7), so `handle_oidc_max_age_param!` / the `prompt=login` path call model methods on a non-model value → `NoMethodError` + `DoubleRenderError`. The dummy app sidesteps this by returning an explicit `nil`, so the suite never exercised it. Good find. 😊

One concern with `next if performed?` though: an unauthenticated user *always* makes the authenticator `performed?` (it redirects to login by design), so this also skips the `prompt=none` path — which OIDC Core §3.1.2.1 requires to return `login_required` without showing any UI. Locally this regresses 6 existing specs (the `prompt=none` "when not logged in" cases under both matching/no-matching-token), since they now get redirected to `/login` instead of the `login_required` error.

What about normalizing the leaked value to `nil` instead of bailing out, so the OIDC handling still runs but with a falsy owner?

```suggestion
            next unless oidc_authorization_request?

            # The configured resource_owner_authenticator redirects unauthenticated
            # users, so `super` may return redirect_to's truthy value rather than a
            # resource owner. Treat that as "no owner" so OIDC handling still runs
            # (prompt=none must yield login_required) without invoking model methods
            # on a non-model value.
            owner = nil if performed?
```

This fixes the same `NoMethodError` / `DoubleRenderError` (the `if owner` / `unless owner` guards now see `nil`), keeps `prompt=none` → `login_required` working, and the full suite stays green (232 examples, 0 failures) — it relies on the same "clear response + re-redirect" path that `handle_oidc_error!` already uses and that the existing specs cover.

Happy to push a repro spec for the leaky-authenticator pattern if useful, since that gap is what hid the bug.

55728: Following up on the `owner = nil if performed?` direction — there's a sibling case in the same code path worth folding into this PR, because **neither** `next if performed?` **nor** `owner = nil if performed?` fully closes it.

`prompt=select_account` hits the same leaky-owner bug. It's not only the leaky case: `select_account` is the one prompt branch with **no `if owner` guard**:

```ruby
when "login"
  reauthenticate_oidc_resource_owner(owner) if owner   # guarded
when "consent"
  if owner                                             # guarded
    clear_oidc_response
    render :new
  end
when "select_account"
  select_account_for_oidc_resource_owner(owner)        # NOT guarded
```

Repro on master (Doorkeeper 5.9, a leaky `current_user || redirect_to(...)` authenticator) with a `select_account_for_resource_owner` block that uses the owner (the realistic case):

```
NoMethodError: undefined method 'id' for an instance of Integer
```

Why it matters for the fix choice:

- `next if performed?` masks it only incidentally — it skips all OIDC handling, the same reason it regresses `prompt=none` → `login_required` (OIDC Core §3.1.2.1).
- `owner = nil if performed?` does **not** fix `select_account`: `select_account_for_oidc_resource_owner(nil)` still runs (no guard), clears the host login redirect, and a host block touching the owner hits `nil.id`. Even a block that ignores the owner still wrongly sends an unauthenticated user to the account chooser instead of login.

So the minimal complete fix is the normalization **plus** giving `select_account` the same guard as `login`/`consent` — a one-liner:

```ruby
when "select_account"
  select_account_for_oidc_resource_owner(owner) if owner
```

That makes `prompt=select_account` with no authenticated user fall through to the host login redirect, consistent with `prompt=login`. Happy to push a repro spec + this guard if useful.

24c02: ooh, this is deeper than i thought! sorry i missed the spec compliance bit there...
please push repro specs and guards if you've got 'em! :-)

55728: Pushed as #281 — includes the `owner = nil if performed?` normalization, the `select_account` guard, and regression specs. Thank you again for the report! 😊

Issue #163, opened 2021-12-05
Is it possible to configure and use Authentication Context Class References?


ChrisMacNaughton:
I was looking at passing information about the level of authentication for the user (MFA status) and wondered if there's support in Doorkeeper / OIDC to handle these claims (ACR and AMR) from [the spec](https://openid.net/specs/openid-connect-core-1_0.html)

55728:
Hi @ChrisMacNaughton 👋 — yes, you can pass ACR and AMR through the existing custom-claims API! Since they're standard top-level ID Token claims ([OIDC Core §2](https://openid.net/specs/openid-connect-core-1_0.html#IDToken)), the `claim` DSL handles them directly:

```ruby
claims do
  claim :acr, response: [:id_token, :user_info], scope: :openid do |resource_owner|
    # OIDC Core §2 — single string, e.g. "urn:mace:incommon:iap:silver",
    # or numeric LoA like "1".."4" per ISO/IEC 29115
    resource_owner.mfa_enabled? ? "2" : "1"
  end

  claim :amr, response: [:id_token, :user_info], scope: :openid do |resource_owner|
    # RFC 8176 — array of strings
    methods = ["pwd"]
    methods << "mfa" if resource_owner.mfa_enabled?
    methods << "otp" if resource_owner.last_login_used_totp?
    methods
  end
end
```

Three notes:

- `response: [:id_token]` is the important bit — custom claims default to `[:user_info]` only, but ACR/AMR are usually expected in the ID Token.
- `scope: :openid` — without this, the claim's scope falls back to `:profile`, so it would silently disappear for clients that only request `openid`.
- `claims_supported` in the discovery document auto-includes your custom claim names. Only `acr_values_supported` (the list of advertised ACR values) isn't generated — that would be a small follow-up PR if you need it.

Does this cover your MFA-status use case? If so, I can send a docs PR to add an ACR/AMR example to the README — it seems to come up often enough to deserve a callout 📝😊

55728:
Hi @ChrisMacNaughton 👋 — doing a bit of issue triage here.

ACR/AMR are supported today through the existing custom-claims `claim` DSL (they're standard top-level ID Token claims), and there's now a README example covering the MFA-status use case, including the `response: [:id_token]` and `scope: :openid` gotchas:

[...]

PR #284, opened 2026-05-23
Document `acr` / `amr` claims in README

55728:
Add a section showing how to expose ACR (Authentication Context Class Reference) and AMR (Authentication Methods References) via the existing `claim` DSL.

Two defaults silently bite when adding standard top-level ID Token claims that aren't in the `STANDARD_CLAIMS` list:

- `response:` defaults to `[:user_info]` only — relying parties usually expect `acr` / `amr` on the ID Token
- `scope:` falls back to `:profile` — clients requesting only `openid` won't see the claim

The example includes both callouts and a note on `claims_supported` vs `acr_values_supported` in the discovery document.

Refs #163.


nbulaj (approved):
LGTM, thanks!

55728:
Thanks for the review! 😊

Issue #159, opened 2021-10-25
Can `.well-known/openid-configuration` return an alternate uri for `jwks_uri`


magic7s:
I'm struggling to find a way to configure an alternate uri for `jwks_uri` in `.well-known/openid-configuration`. I have read [discovery_url_options](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/README.md?plain=1#L166) and found that I can change some things but not the full uri. Any help would be appreciated. 

I need `https://example.com/oauth/discovery/keys` to become `https://example.com/-/jwks`

55728:
Hi @magic7s 👋

I tried @toupeira's snippet against the current dummy app and it actually raises on boot — Rails has refused to let two routes share the same `as:` name [since 4.0](https://github.com/rails/rails/commit/a2b7c0e69d):

```
ArgumentError: Invalid route name, already in use: 'oauth_discovery_keys'
```

But we can still bend the URL helper to a custom path with [`direct`](https://api.rubyonrails.org/classes/ActionDispatch/Routing/Mapper/CustomUrlHelpers.html#method-i-direct) — the discovery controller builds `jwks_uri` from `oauth_discovery_keys_url`, so once we redirect that helper, the discovery document follows:

```ruby
# config/routes.rb
Rails.application.routes.draw do
  use_doorkeeper_openid_connect

  # 1. Mount the custom path under a non-conflicting helper name
  get "/-/jwks",
      to: "doorkeeper/openid_connect/discovery#keys",
      as: :custom_jwks

  # 2. Re-point oauth_discovery_keys_url at the new path
  direct(:oauth_discovery_keys) { |opts| custom_jwks_url(opts) }
end
```

After this:

- `.well-known/openid-configuration` returns `"jwks_uri": "https://example.com/-/jwks"`
- The original `/oauth/discovery/keys` still responds (it's the route doorkeeper-openid_connect mounted), so existing clients aren't broken — handy during a rollover.

Verified against `spec/dummy` on Rails 8.0.5. Happy to send a README PR documenting the pattern if you confirm this covers your use case 📝😊

55728:
Hi @magic7s 👋 — doing a bit of issue triage here.

[...]

PR #285, opened 2026-05-23
Document `jwks_uri` customization via `direct` URL helper

55728:
Add a section showing how to point `jwks_uri` in the discovery document at a custom path using Rails' `direct` URL helper.

The previously suggested approach (redefining a route with `as: :oauth_discovery_keys`) has raised `ArgumentError: Invalid route name, already in use` since [Rails 4.0](https://github.com/rails/rails/commit/a2b7c0e69d). The `direct` helper sidesteps this by overriding only the URL helper without conflicting route names:

```ruby
get "/-/jwks",
    to: "doorkeeper/openid_connect/discovery#keys",
    as: :custom_jwks

direct(:oauth_discovery_keys) { |opts| custom_jwks_url(opts) }
```

Verified against `spec/dummy` on Rails 8.0.5.

Refs #159.


nbulaj (approved):
LGTM ✅  Thanks!

PR #286, opened 2026-05-23
Allow claims to be assigned to multiple scopes

### LINKED ISSUE #125: Is it possible for a claim to have multiple scopes?
I would like to provide an `all_data` scope that acts as a summary of more focused scopes (eg `profile` and `email`), like so:

```
  claims do
    # Profile scope
    claim :given_name, scope: :profile do |user, scopes, access_token|
      user.first_name
    end

    # Email scope
    claim :email, scope: :email do |user, scopes, access_token|
     user.email
    end
  
    # All data scope
    claim :given_name, scope: :all_data do |user, scopes, access_token|
      user.first_name if scopes.exists?(:profile)
    end
    claim :email, scope: :all_data do |user, scopes, access_token|
     user.email
    end
  end
```

From my testing, this currently breaks scope requests for just `profile` or `email`, because the `claims` OpenStruct (from [`ClaimsBuilder`](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/lib/doorkeeper/openid_connect/claims_builder.rb#L18-L36)) uses the claim symbols as keys, thereby overwriting duplicate claim definitions and only storing the claims under the `all_data` scope.

I also can't find in the OIDC spec whether it should be possible to have claims assigned to multiple scopes - does anyone know?

55728:
## Summary

Lets `claim` accept an Array of scopes so the same claim can be exposed under more than one OAuth scope. The original report on #125 was that defining the same claim twice with different `scope:` arguments breaks because `ClaimsBuilder` keys claims by name in an `OpenStruct` (later definitions silently overwrite earlier ones). Following the maintainer's suggestion on the issue, this PR takes the array-form approach instead.

```ruby
claims do
  claim :given_name, scope: [:profile, :all_data] do |user|
    user.first_name
  end

  claim :email, scope: [:email, :all_data] do |user|
    user.email
  end
end
```

A token with `profile` returns `given_name`; a token with `all_data` returns both `given_name` and `email`.

## Changes

- `Claims::Claim#initialize` normalizes `scope:` into a frozen-style `@scopes` array, accepting either a Symbol or an Array. The Standard Claims default-scope fallback (`profile` / `email` / `address` / `phone`) is preserved and now produces a single-element array.
- `Claims::Claim#scope` is kept as a reader returning `@scopes.first` for backward compatibility; `Claims::Claim#scopes` exposes the full Array.
- `ClaimsBuilder.generate` matches a claim when **any** of its scopes is granted by the access token (`claim.scopes.any? { |s| access_token.scopes.exists?(s) }`).
- README gains an array-scope example next to the existing `scope:` documentation.
- New unit specs cover array input, symbolization of String entries, `nil`/empty-array fallback, and the `#scope` back-compat alias. A new integration context in `user_info_spec` exercises array scopes end-to-end.

## Backward compatibility

- `claim :foo, scope: :bar` still works identically.
- `Claim#scope` still returns a single Symbol (the first element of `@scopes`).
- The `scope=` writer that previously came from `attr_accessor :scope` is removed; if any external code mutated it, switching to `claim.instance_variable_set(:@scopes, [...])` or rebuilding the claim is the migration path. This is an internal accessor and is very unlikely to be used externally.

## Out of scope (deliberately)

- Defining the same claim name twice with different scopes is still **not** supported — the claim store is keyed by name. The array form is the documented way to expose a claim under multiple scopes. README is updated to make that explicit.

## Test plan

- [x] `bundle exec rspec` — 252 examples, 0 failures
- [x] `bundle exec rubocop` on touched files — no offenses
- [x] Manual smoke: configure an aggregate scope on a host app and confirm both focused-scope and aggregate-scope tokens emit the expected claims at `/oauth/userinfo` and in the ID token

Closes #125.

### DIFF
--- a/lib/doorkeeper/openid_connect/claims/claim.rb
+++ b/lib/doorkeeper/openid_connect/claims/claim.rb
@@ -4,7 +4,8 @@ module Doorkeeper
   module OpenidConnect
     module Claims
       class Claim
-        attr_accessor :name, :response, :scope
+        attr_accessor :name, :response
+        attr_reader :scopes
 
         # http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims
         # http://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims
@@ -21,15 +22,29 @@ module Doorkeeper
         def initialize(options = {})
           @name = options[:name].to_sym
           @response = Array.wrap(options[:response])
-          @scope = options[:scope].to_sym if options[:scope]
+          @scopes = normalize_scopes(options[:scope])
 
-          # use default scope for Standard Claims
-          @scope ||= STANDARD_CLAIMS.find do |_scope, claims|
+          # use default scope for Standard Claims, fallback to profile
+          @scopes = [default_scope] if @scopes.empty?
+        end
+
+        def scope
+          @scopes.first
+        end
+
+        private
+
+        def normalize_scopes(value)
+          return [] if value.nil?
+
+          scopes = value.is_a?(Array) ? value : [value]
+          scopes.compact.map(&:to_sym)
+        end
+
+        def default_scope
+          STANDARD_CLAIMS.find do |_scope, claims|
             claims.include? @name
-          end.try(:first)
-
-          # use profile scope as default fallback
-          @scope ||= :profile
+          end.try(:first) || :profile
         end
       end
     end
--- a/lib/doorkeeper/openid_connect/claims/claim.rb
+++ b/lib/doorkeeper/openid_connect/claims/claim.rb
@@ -35,10 +35,7 @@ module Doorkeeper
         private
 
         def normalize_scopes(value)
-          return [] if value.nil?
-
-          scopes = value.is_a?(Array) ? value : [value]
-          scopes.compact.map(&:to_sym)
+          Array.wrap(value).compact.map(&:to_sym)
         end
 
         def default_scope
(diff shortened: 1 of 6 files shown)


lib/doorkeeper/openid_connect/claims/claim.rb
```diff
@@ -21,15 +22,29 @@ class Claim
...
+
+        private
 
-          # use profile scope as default fallback
-          @scope ||= :profile
+        def normalize_scopes(value)
+          return [] if value.nil?
+
+          scopes = value.is_a?(Array) ? value : [value]
+          scopes.compact.map(&:to_sym)
```
nbulaj: How about to make it more clean?

```suggestion
          Array.wrap(value).compact.map(&:to_sym)
```

55728:
Thanks for the review, @nbulaj! 😊 I've pushed two commits addressing the feedback:

**1603d75ef0ab5475e137a4015713f7765c117fc6 — `Array.wrap` adoption (your suggestion)**

Replaced the ternary + nil guard with `Array.wrap(value).compact.map(&:to_sym)`. This also aligns with L24 where `@response` already uses `Array.wrap`. Co-authored-by you.

**0211b167b82f3d54769e889dea30c5284cd35bb6 — Copilot feedback**

- Renamed the spec description from "stringifies" → "symbolizes string entries" (it was indeed testing String→Symbol, not the other way around)
- Added a `Claim#scope=` removal note to the CHANGELOG entry

**Re: freezing `@scopes` (Copilot medium)**

I'd prefer to handle this in a separate PR that freezes both `@scopes` and `@response` together — right now neither is frozen, so freezing only `@scopes` would create an inconsistency.

nbulaj (approved):
Hope nobody mutates `scopes` 🙂  Thanks!

PR #287, opened 2026-05-23
Allow `prompt=` handling outside of OIDC requests

### LINKED ISSUE #120: Use prompt=login outside of an oidc request
The prompt=login etc is a nice feature to have even for non OIDC requests. Is there a reason why you restrict it to OIDC requests only?

55728:
## Summary

Adds an opt-in configuration option, `apply_prompt_to_non_oidc_requests` (default `false`), that lets the `prompt` authorization parameter (`none`, `login`, `consent`, `select_account`) take effect on plain OAuth authorization requests that do not include the `openid` scope.

Issue #120 (Jeff Sawatzky, 2021) reported that `prompt=login` and friends are silently ignored outside OIDC requests. Maintainer @toupeira responded that they'd accept a PR adding a single configuration flag because the `prompt` values themselves don't depend on OIDC semantics. This PR implements exactly that.

```ruby
# config/initializers/doorkeeper_openid_connect.rb
  # ...
  apply_prompt_to_non_oidc_requests true
end
```

After enabling, a request like `GET /oauth/authorize?response_type=code&scope=profile&prompt=login&...` triggers `reauthenticate_resource_owner` exactly the same way it would with `scope=openid profile`.

## Changes

- `Config` gains the new `apply_prompt_to_non_oidc_requests` option (default `false`).
- `Helpers::Controller#authenticate_resource_owner!` now branches: OIDC requests get both `handle_oidc_max_age_param!` and `handle_oidc_prompt_param!` as before, non-OIDC requests get only `handle_oidc_prompt_param!` when the new option is enabled, and otherwise are left untouched.
- The existing `oidc_authorization_request?` helper is split: a new `authorization_request_on_authorize_endpoint?` captures the "we are on the authorize endpoint with a valid pre-auth" precondition that both OIDC and the new non-OIDC path need; `oidc_authorization_request?` keeps its previous public-looking shape (still requires the `openid` scope).
- Spec coverage: controller-level specs for the flag-on path covering `login`, `none`, `consent`, `select_account`, and an unknown prompt value; plus a guard test confirming `max_age` stays OIDC-only even when the flag is enabled. New `config_spec.rb` cases cover the default and overridden values.
- README documents the new option next to the existing optional settings.

## Design notes

- **`max_age` deliberately stays OIDC-only.** It's defined by OIDC Core §3.1.2.1 and is conceptually a separate parameter from `prompt`; @toupeira's comment on the issue specifically scoped the flag to `prompt`.
- **OIDC error codes (`login_required`, `consent_required`, `account_selection_required`) leak into OAuth flows when the flag is enabled.** This is intentional and is the natural consequence of running `handle_oidc_prompt_param!` on the request. Clients that already integrate with this gem are OIDC-aware, so the leakage is acceptable.
- **Default off** means zero behavior change for existing installations.
- **`prompt=create` continues to be a no-op pass-through** on both OIDC and non-OIDC paths.

## Test plan

- [x] `bundle exec rspec` — 253 examples, 0 failures
- [x] `bundle exec rubocop` on touched files — no offenses
- [x] Manual smoke: with `apply_prompt_to_non_oidc_requests true`, confirm `GET /oauth/authorize?scope=profile&prompt=login&...` redirects through `reauthenticate_resource_owner`, and that `scope=profile&prompt=none` (no current user) returns `login_required` to the registered `redirect_uri`

Closes #120.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - [#285] Document custom `jwks_uri` path pattern in README — show how to advertise a non-default path in the discovery document using Rails' `direct` URL helper
 - [#283] Support multiple signing keys in the JWKS response — `signing_key` now also accepts an array (and callables returning an array). The first entry is the active key used to sign new ID tokens; the remaining entries are published in the JWKS so clients can still validate tokens signed with a retired key during a rotation window. Single-value and callable forms continue to work unchanged
 - [#286] Allow claims to be assigned to multiple scopes via `scope: [:profile, :all_data]` — the claim is returned whenever the access token grants any of the listed scopes. **Note:** the previously implicit `Claim#scope=` writer (from `attr_accessor :scope`) is no longer provided; rebuild the claim instead of mutating it
+- [#287] Add `apply_prompt_to_non_oidc_requests` option to honor the `prompt` parameter on plain OAuth requests that do not include the `openid` scope
 
 ## v1.9.0 (2026-03-16)
 
--- a/README.md
+++ b/README.md
@@ -246,6 +246,20 @@ The following settings are optional:
     end
     ```
 
+- `apply_prompt_to_non_oidc_requests`
+  - Whether to honor the `prompt` authorization parameter (`none`, `login`, `consent`, `select_account`) on plain OAuth requests that do not include the `openid` scope.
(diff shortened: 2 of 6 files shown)


nbulaj (approved):
LGTM ✅

PR #288, opened 2026-05-23
Document offline_access scope recipe in README

### LINKED ISSUE #133: `offline_access` scope support
The [offline_access scope](https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess) defined in the OIDC specification appears to be completely unsupported as yet.

So as to not break peoples' existing oidc flows it would probably be prudent to put this in behind a config option.

I *think* for the application I'm working on it's not security-critical to have this feature, but if I find out otherwise I'll likely end up implementing it in Q1 / Q2 2021.

55728:
### Summary

Add a recipe showing how to wire Doorkeeper's `use_refresh_token` block with `offline_access` scope filtering. The basic flow requires no gem changes — just `optional_scopes` + a scope-aware `use_refresh_token` block.

Includes a note that this does not automatically enforce OIDC Core §11's strict requirements (e.g. MUST ignore `offline_access` when `prompt=consent` is absent).

### Context

Issue #133 (open since 2020) asked for `offline_access` support. Triage confirmed the existing Doorkeeper primitives already cover the use case. The original reporter [confirmed](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/133#issuecomment-4525540674) the workaround is sufficient.

Closes #133.


nbulaj (approved):
Thanks! 👍 

Maybe we need to full-fill a Wiki here or introduce Git book docs as we have for Doorkeeper 👀 

Since README can grow enough with all the possible scenarios / examples

Issue #192, opened 2023-03-02
Support multiple devise models


bilouw:
Hi !

My Rails app use devise and doorkeeper. My app have 2 (in fact 3, but let begin with 2) devise model: User and Customer.

Doorkeeper by default handle only one devise model. By that, i mean that generators and example are based on use case with only one devise model.

With some controllers override, we can make doorkeeper handle multiples devise models.

Now, i have the same problem with this gem. I found a workaround (the same thing i did with doorkeeper), but i want to know if there is a legacy way to do that, or if an PR could be done to handle that.

Let me explain: When i want to access the /.well-known/openid-configuration url, the DiscoveryController try to build the json object with some hardcoded helpers: 

<img width="879" alt="image" src="https://user-images.githubusercontent.com/7089641/222516875-1ddc03fd-29ff-493e-a10a-c4d37b12b886.png">

Because i have two devise model (in two namespaces), these helpers methods does not exists for me.
What i did: Override the DiscoveryController in each of my namespaces, then override the provider_response methods to correct the helpers methods.

For now, that work !

But want to know if anybody have already handle this use case or if there is a best practice to handle this case.

Thanks !

55728:
Hi @bilouw, thank you for the detailed write-up and for sharing your workaround.

As a collaborator: a quick clarification first — Doorkeeper itself isn't strictly tied to a single resource owner model. The `resource_owner_authenticator` block is model-agnostic, so a host app can dispatch to `current_user || current_customer` within a single Doorkeeper mount, and claims / userinfo / ID token generation flow naturally from whatever the block returns.

The actual sticking point in your case is narrower: `DiscoveryController#provider_response` calls named route helpers like `oauth_authorization_url` directly ([source](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/c4260ce88d522c79305031e21124e2f255f47193/app/controllers/doorkeeper/openid_connect/discovery_controller.rb)), which becomes ambiguous when the engine is mounted under multiple namespaces.

The existing [`discovery_url_options`](https://github.com/doorkeeper-gem/doorkeeper-openid_connect#configuration) setting (added in #126) covers per-endpoint `host` / `protocol` overrides, but doesn't let you swap which named helper is invoked, so it doesn't fully cover the multi-mount case. Your per-namespace controller override is currently the most idiomatic way to handle this, and it's a supported extension pattern for Rails engines.

[...]

PR #126, opened 2020-09-24
Add discovery_url_options option for discovery endpoints URL generation.

phlegx:
Like described from @steti in PR #90, this is a re-work to be able to configure all discovery endpoint URL generations.

Example:

```ruby
Doorkeeper::OpenidConnect.configure do
  # ...

  discovery_url_options do |request| 
    {
      authorization: { host: 'alternate-authorization.host' },
      token: { host: 'alternate-token.host' },
      revocation: { host: 'alternate-revocation.host' },
      introspection: { host: 'alternate-introspection.host' },
      userinfo: { host: 'alternate-userinfo.host' },
      jwks: { host: 'alternate-jwks.host', protocol: request.ssl? ? :https : :http }
    }
  end

  # ...
end
```

or for only one endpoint:

```ruby
Doorkeeper::OpenidConnect.configure do
  # ...

  discovery_url_options do
    {  authorization: { host: 'alternate-authorization.host' } }
  end

  # ...
end
```

PR #291, opened 2026-05-30
Document multi-namespace mount pattern for multiple resource owner models

55728:
## Summary

Docs-only. Adds a **Routes** subsection describing how to mount Doorkeeper OpenID Connect under multiple namespaces (e.g. separate `User` / `Customer` Devise models), and the one caveat that involves: the discovery document.

## Why

`DiscoveryController#provider_response` builds the published endpoints by calling named route helpers (`oauth_authorization_url`, …) directly. `discovery_url_options` (#126) can override the `host` / `protocol` / `port` of those URLs, but not *which* named helper is resolved — so under multiple mounts every namespace's discovery document would point at the same endpoints. The note documents the idiomatic fix: a per-namespace `DiscoveryController` subclass that re-points the helpers at its own routes.

## Notes

- No code changes — README only.
- Refs #192. This is documented as the supported extension pattern for now; whether to add first-class multi-mount support is left as a separate decision (cc @nbulaj).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Please add here
 - [#294] Drop stale `Metrics/ClassLength` and `Metrics/BlockLength` overrides from `.rubocop_todo.yml`
 - [#293] Drop `Naming/VariableNumber` from `.rubocop_todo.yml` and normalise test variable names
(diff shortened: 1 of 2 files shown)


nbulaj (approved):
I love it 🤟

PR #292, opened 2026-05-30
Drop formatting cops from rubocop_todo, align trailing commas with upstream

55728:
Remove six auto-correctable formatting suppressions from `.rubocop_todo.yml` and fix the underlying offenses.

- **Reproducible:** the code diff is entirely `bundle exec rubocop -A` — no hand-written changes. Resetting code files to `master` and re-running `-A` reproduces the result byte-for-byte.
- **Upstream-aligned:** trailing-comma cops use `consistent_comma` to match doorkeeper 5.9.1. Layout cops follow RuboCop defaults (same as upstream).

rubocop: 92 files, 0 offenses. rspec: 282 examples, 0 failures.

### DIFF
--- a/lib/doorkeeper/oauth/id_token_token_response.rb
+++ b/lib/doorkeeper/oauth/id_token_token_response.rb
@@ -5,10 +5,10 @@ module Doorkeeper
     class IdTokenTokenResponse < IdTokenResponse
       def body
         super.merge({
-          access_token: auth.token.token,
-          token_type: auth.token.token_type,
-          expires_in: auth.token.expires_in_seconds
-        })
+                      access_token: auth.token.token,
+                      token_type: auth.token.token_type,
+                      expires_in: auth.token.expires_in_seconds,
+                    })
       end
     end
   end
--- a/lib/doorkeeper/oauth/id_token_token_response.rb
+++ b/lib/doorkeeper/oauth/id_token_token_response.rb
@@ -5,10 +5,10 @@ module Doorkeeper
     class IdTokenTokenResponse < IdTokenResponse
       def body
         super.merge({
-                      access_token: auth.token.token,
-                      token_type: auth.token.token_type,
-                      expires_in: auth.token.expires_in_seconds,
-                    })
+          access_token: auth.token.token,
+          token_type: auth.token.token_type,
+          expires_in: auth.token.expires_in_seconds,
+        })
       end
     end
   end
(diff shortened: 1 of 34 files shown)


lib/doorkeeper/oauth/id_token_token_response.rb
```diff
@@ -5,10 +5,10 @@ module OAuth
     class IdTokenTokenResponse < IdTokenResponse
       def body
         super.merge({
-          access_token: auth.token.token,
-          token_type: auth.token.token_type,
-          expires_in: auth.token.expires_in_seconds
-        })
+                      access_token: auth.token.token,
```
nbulaj: Hm, this one looks like we need to change the rule. Hashes/arguments placed to the right (especially when have long sentences) are hard to read. Previous implementation is something I'd like to see in the code base 👀  WDYT?

spec/controllers/doorkeeper/authorizations_controller_spec.rb
```diff
@@ -674,10 +674,10 @@ def authorize_with_session!(session_attrs, params = {})
         authorize! max_age: 10, prompt: "none", state: "somestate"
 
         expect(response).to redirect_to build_redirect_uri({
-          "error" => "login_required",
-          "error_description" => "The authorization server requires end-user authentication",
-          "state" => "somestate",
-        })
+                                                             "error" => "login_required",
+                                                             "error_description" => "The authorization server requires end-user authentication",
```
nbulaj: This is a good example of the above one, just see the diff

Image

Even on my 24" display it looks odd 👀

lib/doorkeeper/oauth/id_token_token_response.rb
55728: Agreed, I'll switch to `Layout/FirstHashElementIndentation: consistent` — pushing a fix shortly 👀✨

spec/controllers/doorkeeper/authorizations_controller_spec.rb
55728: Your display can breathe easy now 😄

55728:
Pushed a follow-up commit to clean up the cramped `,]` / `,)` closers that `consistent_comma` introduced. Where line length allowed (specs), collapsed to a single line; otherwise expanded to one-element-per-line with the closer on its own line — matching the style already used in sibling code. These are hand-edited, not `rubocop -A`.

nbulaj (approved):

55728:
Thank you for the review 😊

PR #303, opened 2026-06-04
execute account selection even without owner

### LINKED ISSUE #302: `select_account` now doesn't trigger account selection when no `owner`
ref. https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/281

when IdP handles account selection, there can be candidates all are not authenticated but stored in the session.
so that users can select account, then authenticate on the account.

however #281 stopped executing account selection when no `owner`

nov:
fixes https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/302

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - Please add here
+- [#303] execute account selection even without owner
 
 ## v1.10.1 (2026-06-03)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Unreleased
 
 - Please add here
-- [#303] execute account selection even without owner
+- [#303] execute account selection even without owner, and `select_account_for_resource_owner` can now receive `nil` as the first argument.
 - [#304] allow handle auth_time per grant
 
 ## v1.10.1 (2026-06-03)
--- a/lib/doorkeeper/openid_connect/helpers/controller.rb
+++ b/lib/doorkeeper/openid_connect/helpers/controller.rb
@@ -125,7 +125,7 @@ module Doorkeeper
                 render :new
               end
             when "select_account"
-              select_account_for_oidc_resource_owner(owner) if owner
+              select_account_for_oidc_resource_owner(owner)
             when "create"
               # NOTE: not supported, but not raise error.
             else
--- a/lib/generators/doorkeeper/openid_connect/templates/initializer.rb
+++ b/lib/generators/doorkeeper/openid_connect/templates/initializer.rb
@@ -76,9 +76,11 @@ Doorkeeper::OpenidConnect.configure do
     # redirect_to new_user_session_url
   end
 
-  select_account_for_resource_owner do |resource_owner, return_to|
+  select_account_for_resource_owner do |resource_owner_or_nil, return_to|
     # Example implementation:
-    # store_location_for resource_owner, return_to
+    # if resource_owner_or_nil
+    #   store_location_for resource_owner_or_nil, return_to
+    # end
     # redirect_to account_select_url
   end
 
--- a/spec/controllers/doorkeeper/redirecting_resource_owner_authenticator_spec.rb
+++ b/spec/controllers/doorkeeper/redirecting_resource_owner_authenticator_spec.rb
@@ -61,13 +61,13 @@ describe Doorkeeper::AuthorizationsController, type: :controller do
     # like the other unauthenticated prompts.
     it "prompt=select_account: falls through to the login redirect without raising" do
       Doorkeeper::OpenidConnect.configure do
-        select_account_for_resource_owner do |resource_owner, _return_to|
-          redirect_to "/accounts?uid=#{resource_owner.id}"
+        select_account_for_resource_owner do |_resource_owner, _return_to|
+          redirect_to "/accounts"
         end
       end
 
       expect { authorize!(prompt: "select_account") }.not_to raise_error
-      expect(response).to redirect_to("/login")
+      expect(response).to redirect_to("/accounts")
     end
 
     # OpenID Connect Core 1.0 §3.1.2.1: with prompt=none and no authenticated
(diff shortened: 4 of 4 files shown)


55728 (commented):
Thanks for the detailed explanation in #302 @nov — the Google account-selection screenshot makes the use case very clear. I agree that `prompt=select_account` should be able to work before authentication.

However, I'm concerned that simply removing the `if owner` guard introduces a breaking change for existing host apps. The current `select_account_for_resource_owner` callback contract implies a non-nil `resource_owner` — in fact the gem's own generated initializer template passes `resource_owner` to `store_location_for`, so the non-nil contract is baked into what we ship. Many implementations will dereference it:

```ruby
select_account_for_resource_owner do |resource_owner, return_to|
  redirect_to account_select_url(uid: resource_owner.id)
end
```

I verified locally that this pattern raises `NoMethodError` (undefined method `id` for nil) on this branch, while it works fine on master.

As an alternative, what do you think about an **opt-in config** approach? This mirrors the design of `auth_time_from_access_token` in #304 — default preserves existing behavior, opt-in enables the new flow:

```ruby
when "select_account"
  if owner
    select_account_for_oidc_resource_owner(owner)
  elsif Doorkeeper::OpenidConnect.configuration.allow_select_account_without_owner
    select_account_for_oidc_resource_owner(nil)
  end
```

Of course, the final call on whether to gate this behind a config or document the nil-contract change is up to @nbulaj — happy to go either way. Would you be open to updating this PR in that direction?

nov:
the addition of `if owner` is active only 3 weeks.
https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/281

that code was working without it for long.

why do you care about it?

55728:
Good question! The `if owner` guard was added in #281 as part of fixing the leaky-authenticator issue from #280 — when `resource_owner_authenticator` redirects (e.g. for an unauthenticated user), the `redirect_to` return value (an Integer in Rails 8) leaked through as `owner`, causing `NoMethodError` in downstream blocks. The guard ensured that only a real model object reaches the callback.

Fair point on the `uid` example — that's just one possible pattern. The broader concern is about the callback contract changing from "always receives a model" to "may receive nil", which could break any implementation that calls a method on the argument.

That said, I've already deferred the final call to @nbulaj, so happy to wait for his take on the right approach.

nov:
this might be enough?
https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/303/changes/e9acc912b374dae5e991de0ce990e9c6ee7a165b

55728:
Thanks for the update — the `resource_owner_or_nil` naming in the template makes the contract change explicit. Let's see what @nbulaj thinks.

55728:
After digging into the history a bit more, I realized the core issue is that `if owner` was conflating two different scenarios:

**A) Leaky authenticator** (`performed?` = true): the authenticator redirected, and `owner = nil if performed?` (added in #281) normalizes the leaked value to `nil`. Removing the guard here means `nil` reaches the callback for the first time ever — pre-1.10.0 passed the raw `redirect_to` return value (Integer 302 on Rails 8 / a String body on Rails 7), not `nil`.

**B) Clean nil return** (`performed?` = false): the authenticator deliberately returned `nil` without redirecting — this is the #302 use case. Pre-1.10.0 this worked, and #281's guard (added to fix the #280 leaky-authenticator crash) was too broad and disabled this path too.

So rather than guarding on `owner`, we can guard on `performed?` to distinguish A from B:

```ruby
when "select_account"
  if owner
    select_account_for_oidc_resource_owner(owner)
  elsif !performed?
    select_account_for_oidc_resource_owner(nil)
  end
```

This way:
- Scenario A (leaky, `performed?` true): falls through to the existing login redirect ✅
- Scenario B (clean nil, `performed?` false): callback fires with `nil` owner — #302 fixed ✅
- Authenticated: works as before ✅

I've verified locally that this passes the full suite (296 examples, 0 failures) including both scenarios. Would this approach work for your use case?

nov:
I want the visiter redirect to account selector even not authenticated.

I'm on a vacation for a whole, and only having iPhone right now. so will check the details after that.

55728:
Thanks @nov — I'm in favor of this direction and would like to see it merged. 😊

I initially proposed a config gate and then an `elsif !performed?` guard, but after digging deeper I've concluded neither is needed — the normalization (`owner = nil if performed?`) already closes #280, so the unconditional call is safe.

I went back through #280 → #281 → #302 to make sure dropping the guard doesn't reintroduce anything. My read:

**This doesn't reintroduce #280.** The fix that actually addresses #280 is the normalization in `authenticate_resource_owner!` (`owner = nil if performed?`), and this PR leaves it untouched. With it in place, the callback can only ever receive a real resource owner or `nil` — the leaked Integer/String values from #280 stay structurally impossible. The `if owner` guard was an extra layer I added in #281, and #302 shows it rested on a wrong assumption: it treated `nil` as if it could only mean "something leaked", but `nil` is also a legitimately unauthenticated visitor — exactly the state `prompt=select_account` exists to handle (OIDC Core §3.1.2.1 defines it as accounts the user *might* have current sessions for; it doesn't presuppose an established authentication). Notably, #280 never asked for the call to be suppressed on `nil` — the normalization was the requirement there, and it stays.

**The breaking surface is small, and it fails closed.** The guard landed ~3 weeks ago and has only been in a release since 1.10.0 (June 1). A host block that dereferences the owner without a nil check would 500 on the unauthenticated path — loud, not silent, and never in the direction of an authorization incorrectly succeeding. A block that simply ignores `nil` still falls through to `AccountSelectionRequired`. If an edge case we haven't anticipated does surface, it will announce itself and we can address it then.

**Follow-up I'm happy to handle** (separate PR, unless you'd prefer it folded in here):
1. Regression specs for the new contract — in particular asserting the callback receives `nil` (not a leaked non-model value) under a redirecting authenticator. That's a stronger form of the #280 protection than the old "callback is not called" spec.
2. CHANGELOG note documenting the behavior change ("the block may now receive `nil` for unauthenticated visitors") with a short upgrade hint — deferring to @nbulaj on whether this warrants a minor bump.
3. README note for `select_account_for_resource_owner` mentioning that the block may receive `nil` (the initializer template already covers this nicely — thanks for e9acc912 👍).

One tiny thing in this PR: the spec description (and the comment above it) still reads "falls through to /login" while the assertion now expects `/accounts`. Happy to fold that touch-up into the follow-up too, if you'd rather not push again.

nov:
OK, added some spec & updated the CHANGELOG.

55728 (approved):
Thanks for the thorough updates @nov — the `raise unless resource_owner_or_nil.nil?` assertion in the spec is a really strong regression guard. LGTM 👍

nov:
thanks a lot!

PR #304, opened 2026-06-04
allow handle auth_time per grant

nov:
currently, `auth_time` in ID Token is given via `auth_time_from_resource_owner`.

it's fine in most cases, but theoretically it's not always the same with the time the authorization grant is generated.
(users can switch sessions during authorization code issued and authorization code exchanged)

thus, in my application, I'm storing `auth_time` in a custom record representing authentication context on authorization grant insurance.

the new config `auth_time_from_access_token` allows get `auth_time` from each tokens, not from users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - Please add here
+- [#304] allow handle auth_time per grant
 
 ## v1.10.1 (2026-06-03)
 
--- a/lib/doorkeeper/openid_connect/config.rb
+++ b/lib/doorkeeper/openid_connect/config.rb
@@ -54,6 +54,7 @@ module Doorkeeper
       }
 
       option :auth_time_from_session, default: nil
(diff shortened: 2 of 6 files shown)


55728 (commented):
Thanks for this addition, @nov! 😊 The per-grant `auth_time` approach makes a lot of sense — it's the most precise way to handle the session-switch scenario you described.

One small thing I noticed in `spec/lib/config_spec.rb`:

```ruby
describe "auth_time_from_access_token" do
  it "defaults to nil" do
    described_class.configure {}

    expect(subject.auth_time_from_session).to be_nil
    #                    ^^^^^^^^^^^^^^^^
    # Should this be auth_time_from_access_token?
  end
end
```

Both are `nil` so the test passes, but it's not actually asserting the new option's default.

nov:
sorry, fixed the typo because of old style copy & paste :p

55728 (approved):
Makes sense, thanks 😊

Issue #289, opened 2026-05-25
Migrate documentation from README to Wiki


55728:
## Summary

As the project matures and more recipes/guides are added to the README, it may be time to consider moving documentation to a dedicated platform — either the GitHub Wiki or GitBook (as Doorkeeper itself uses at https://doorkeeper.gitbook.io/guides/).

## Context

In #288, @nbulaj noted:

> Maybe we need to full-fill a Wiki here or introduce Git book docs as we have for Doorkeeper 👀
> Since README can grow enough with all the possible scenarios / examples

The README has been growing with configuration examples and recipes (e.g. the `offline_access` recipe added in #288). As more OIDC-specific scenarios are documented — claims customization, session management, `prompt` parameter handling, etc. — a single README will become harder to navigate.

### 1. GitHub Wiki

- **Pros:** Zero setup, built into the repo, easy for contributors to edit
- **Cons:** No PR-based review workflow, not versioned alongside releases, limited formatting/structure

### 2. GitBook (aligned with Doorkeeper)

- **Pros:** Consistent with the Doorkeeper ecosystem (`doorkeeper.gitbook.io`), rich navigation/search, versioning support
- **Cons:** Requires separate hosting setup, another platform to maintain

### 3. Keep README (status quo)

- **Pros:** Single source of truth, discoverable, no extra tooling
- **Cons:** Will continue to grow; harder to find specific topics as recipes accumulate

## Proposal

I don't have a strong preference yet — opening this to gather input from maintainers and the community. Some questions to consider:

- Is there an existing GitBook organization/workspace under `doorkeeper-gem` that this project could join?
- Would a lightweight approach (e.g. a `docs/` folder with markdown files) be a better first step before committing to an external platform?

Happy to help with the migration work once a direction is decided.

nbulaj:
Hello. I';d say we can start from the Wiki for this repo. It's still allows external contributions and saves data as a markdown so when docs grew up we can easily move them to GitBook for example. Saving docs in a separate dir can be non SEO-friendly and not as transparent and common as a GitHub Wiki. WDYT?

55728:
Thanks for the direction! Wiki sounds like the right starting point — Markdown-based, contributor-friendly, and easy to migrate later if we outgrow it.

I'll start moving content from the README into Wiki pages incrementally. 👀✨

PR #305, opened 2026-06-07
Document `auth_time_from_access_token` config option in README

55728:
## Summary

#304 added the `auth_time_from_access_token` config option, allowing `auth_time` to be tracked **per grant** (derived from the access token) instead of **per user** (derived from the resource owner). That PR documented the option in the initializer template and added a CHANGELOG entry, but the **README option list was not updated**.

This PR adds the missing README documentation.

## Changes

- **README.md** — add an `auth_time_from_access_token` entry alongside `auth_time_from_resource_owner` / `auth_time_from_session`, including a usage example.
- **CHANGELOG.md** — add an `## Unreleased` entry for this docs change.

## Documentation notes

The README entry intentionally calls out two points verified against the implementation:

1. **Precedence** — when `auth_time_from_access_token` is configured, it takes precedence over `auth_time_from_resource_owner` for populating the `auth_time` claim on the ID Token (`lib/doorkeeper/openid_connect/id_token.rb`).
2. **Scope limitation** — it **only** affects the ID Token `auth_time` claim. It is **not** used for `max_age` enforcement, which still resolves auth_time via `auth_time_from_session` / `auth_time_from_resource_owner` in the controller helper (`lib/doorkeeper/openid_connect/helpers/controller.rb`). This is highlighted to avoid the misreading that switching to per-grant `auth_time` also makes `max_age` per-grant.

## Related

- Follows up #304 (the feature PR that introduced `auth_time_from_access_token`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Please add here
 - [#304] allow handle auth_time per grant
+- [#305] Document the `auth_time_from_access_token` config option in the README (per-grant `auth_time`), clarifying that it only affects the ID Token `auth_time` claim and not `max_age` enforcement
 
 ## v1.10.1 (2026-06-03)
 
--- a/README.md
+++ b/README.md
@@ -180,6 +180,16 @@ The following settings are optional, but recommended for better client compatibi
       session[:auth_time]
     end
     ```
+- `auth_time_from_access_token`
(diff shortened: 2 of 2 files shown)


55728:
Once this is merged, the README will cover all main features — so I'll start migrating documentation to the Wiki as part of #289 👀✨

nbulaj (approved):

55728:
Thank you for the review!

PR #307, opened 2026-06-09
Fix `bundle exec rake server` for the test application

55728:
### Problem

The [Development section](https://github.com/doorkeeper-gem/doorkeeper-openid_connect#development) of the README documents `bundle exec rake server` to run the local engine server, but it currently has three issues:

1. It fails with `Could not find a server gem` because no server gem is listed in the Gemfile.
2. After adding a server gem, accessing `http://localhost:3000/` raises `ActionController::RoutingError (uninitialized constant DummyController)` because `routes.rb` references `root 'dummy#index'` but no `DummyController` exists.
3. The `server` Rake task inherits `ENV["RAILS_ENV"] ||= "test"` from the top of the Rakefile, so the test application runs under the `test` environment instead of `development`.

### Changes

- Add `puma` to the `Gemfile` under `group :development`
- Add `DummyController#index` that redirects to `/.well-known/openid-configuration`
- Preserve any user-specified `RAILS_ENV` and default to `development` in the `server` Rake task

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-ENV["RAILS_ENV"] ||= "test"
-
 require "bundler/gem_tasks"
 require "rspec/core/rake_task"
 
@@ -12,6 +10,7 @@ task test: :spec
 
 desc "Generate and run migrations in the test application"
 task :migrate do
+  ENV["RAILS_ENV"] ||= "test"
   Dir.chdir("spec/dummy") do
     system("bin/rails generate doorkeeper:openid_connect:migration")
     system("bin/rake db:migrate")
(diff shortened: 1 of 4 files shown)


Rakefile
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+USER_RAILS_ENV = ENV["RAILS_ENV"]
 ENV["RAILS_ENV"] ||= "test"
```
nbulaj: BTW I beleive it's a legacy code, do we really need it? Rspec should set RAILS_ENV in the rail_helper/spec_helper so we don't need to patch RAILS_ENV when running rake tasks

55728: Good point! 👀✨ I've removed the top-level `ENV["RAILS_ENV"] ||= "test"` entirely and scoped it per task instead:

- `migrate` task → defaults to `"test"`
- `server` task → defaults to `"development"`

Force-pushed. Please take another look when you get a chance.

nbulaj (approved):

PR #308, opened 2026-06-12
Wire `openid_request` from the AccessGrant mixin instead of `on_load(:active_record)`

### LINKED ISSUE #306: NameError: uninitialized constant when using namespaced custom models since   v1.10.0 — needs same fix as doorkeeper#1830
Hi. I encountered an issue while using this Gem v1.10.1.
Since v1.10.0, applications using namespaced custom models crash on Rails initialization:
`NameError: uninitialized constant Auth::ApplicationRecord`

This is a regression from v1.9.0, and the root cause is the same issue that doorkeeper fixed in doorkeeper#1830 (released as doorkeeper v5.9.2).

## Environment

- doorkeeper-openid_connect: 1.10.1 (not reproducible on 1.9.0)
- doorkeeper: 5.9.2
- Rails: 8.1.x / Ruby: 3.4.x

## Background

```ruby
# config/initializers/doorkeeper.rb
access_grant_class 'Auth::OAuthAccessGrant'

# app/models/auth/oauth_access_grant.rb
module Auth
  class OAuthAccessGrant < ApplicationRecord
    include Doorkeeper::Orm::ActiveRecord::Mixins::AccessGrant
  end
end
```

Root Cause

The problem is a chain of three changes:

1. doorkeeper v5.9.0 (PR #1804): Wrapped model loading in on_load(:active_record) to defer it.
2. doorkeeper-openid_connect v1.10.0 (PR #241): Followed suit and wrapped run_hooks in on_load(:active_record).
3. doorkeeper v5.9.2 (PR #1830): Discovered that on_load(:active_record) itself was the problem and removed it entirely.

doorkeeper-openid_connect v1.10.0 was a correct response to step 1, but step 3 invalidated the approach. doorkeeper-openid_connect still uses on_load(:active_record) even though doorkeeper itself has abandoned it.

The specific failure happens like this:

## Rails boot sequence
class ApplicationRecord < ActiveRecord::Base   # (1) evaluates ActiveRecord::Base
                                               # (2) run_load_hooks(:active_record) fires HERE
                                               # (3) ::ApplicationRecord is registered HERE (this timing is too late)

At step (2), run_hooks calls Doorkeeper.config.access_grant_model ('Auth::OAuthAccessGrant'.constantize), which loads
auth/oauth_access_grant.rb.
That file references ApplicationRecord inside module Auth, but `::ApplicationRecord` is not yet registered (step 3 hasn't happened), causing the NameError.

[doorkeeper#1830](https://github.com/doorkeeper-gem/doorkeeper/pull/1830) solved this by removing on_load(:active_record) from run_hooks and moving the setup into each Mixins::* included block instead.

55728:
Fixes #306

## Problem

Since v1.10.0, apps using a **namespaced custom access grant model** crash on Rails boot:

```ruby
# config/initializers/doorkeeper.rb
Doorkeeper.configure do
  access_grant_class "Auth::OAuthAccessGrant"
end

# app/models/auth/o_auth_access_grant.rb
module Auth
  class OAuthAccessGrant < ApplicationRecord
    include Doorkeeper::Orm::ActiveRecord::Mixins::AccessGrant
  end
end
```

```
NameError: uninitialized constant Auth::ApplicationRecord (NameError)
```

## Root cause

v1.10.0 (#241) wrapped the access-grant prepend in `ActiveSupport.on_load(:active_record)`, following doorkeeper's [#1804](https://github.com/doorkeeper-gem/doorkeeper/pull/1804). doorkeeper itself later reverted that approach in [#1830](https://github.com/doorkeeper-gem/doorkeeper/pull/1830) (v5.9.2, see [doorkeeper#1828](https://github.com/doorkeeper-gem/doorkeeper/issues/1828)): the `:active_record` load hook fires while `ActiveRecord::Base` is first being loaded — typically mid-evaluation of `class ApplicationRecord < ActiveRecord::Base`. Constantizing the configured grant class from inside the hook at that moment resolves `ApplicationRecord` relative to the model's namespace before `::ApplicationRecord` has been registered, raising the `NameError`.

## Fix

Follow the same direction as doorkeeper [#1830](https://github.com/doorkeeper-gem/doorkeeper/pull/1830): wire from the mixin, not from a deferred load hook.

- Drop the `run_hooks` / `initialize_models!` overrides and the `on_load(:active_record)` block entirely.
- Prepend a small extension onto the singleton class of `Doorkeeper::Orm::ActiveRecord::Mixins::AccessGrant`. Its `included` callback prepends `Doorkeeper::OpenidConnect::AccessGrant` (which adds the `openid_request` association) onto every model that includes the mixin — the default `Doorkeeper::AccessGrant` as well as any custom model.
- This runs at the host model's own load time, with the model class handed in by Ruby as `base`. Nothing constantizes the configured grant class anymore, so the re-entrant load window is gone. The string `class_name:` keeps the association target resolved lazily by ActiveRecord.
- The prepend is guarded with `base.is_a?(Class)`: if the mixin is included into an intermediate `ActiveSupport::Concern`, the hook first fires with that module as `base`; the concern defers the include, so the hook fires again with the actual model class.
- The legacy `active_record_options[:establish_connection]` handling moves into the `OpenidRequest` mixin's `included` block (guarded with `respond_to?`; a no-op on doorkeeper 5.9.x, which no longer exposes `active_record_options`).
- The old `doorkeeper < 5.5` fallback branch (`Doorkeeper::AccessGrant.prepend ...`) is removed — the gemspec already requires `doorkeeper >= 5.5`, where `Mixins::AccessGrant` exists.

### Behavior note

Previously only the configured `Doorkeeper.config.access_grant_model` was prepended; now every model that includes doorkeeper's `Mixins::AccessGrant` gains the association. For documented setups these are equivalent, and the new behavior matches the include semantics doorkeeper itself uses since #1830.

## Testing

- New regression specs in `spec/lib/orm/active_record_spec.rb`:
  - the default access grant model has the `openid_request` association;
  - a namespaced custom model that includes doorkeeper's mixin (mirroring the #306 reporter's setup) is wired with the association. This spec fails under the old `on_load` implementation, since the custom model is wired only via the mixin hook.
  - The boot-time `NameError` itself cannot be reproduced inside the suite (ActiveRecord is already loaded by then), so the specs pin the wiring mechanism instead; the original crash was reproduced and verified fixed manually in a Rails app using the #306 reporter's setup.
- Full suite: **293 examples, 0 failures** (from a clean test DB), RuboCop clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 - Please add here
+- [#308] Fix `NameError: uninitialized constant Auth::ApplicationRecord` on boot when using a namespaced custom access grant model (e.g. `Auth::OAuthAccessGrant < ApplicationRecord`). Since v1.10.0 ([#241]) the `openid_request` association was wired inside an `ActiveSupport.on_load(:active_record)` block, which fires while `ActiveRecord::Base` is first loaded and constantizes the grant model too early. The association is now added from Doorkeeper's `AccessGrant` mixin `included` callback — at the model's own load time, without constantizing — mirroring the fix doorkeeper made in [#1830](https://github.com/doorkeeper-gem/doorkeeper/pull/1830) ([#306](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/306))
 
 ## v1.10.2 (2026-06-22)
 
--- a/lib/doorkeeper/openid_connect/orm/active_record.rb
+++ b/lib/doorkeeper/openid_connect/orm/active_record.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "active_support/lazy_load_hooks"
-
 module Doorkeeper
   module OpenidConnect
     autoload :AccessGrant, "doorkeeper/openid_connect/orm/active_record/access_grant"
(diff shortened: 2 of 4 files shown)


nbulaj (commented):
Looks a little bit tricky and like smth which will cause new issues create 😄  let me sleep on it to see our next steps

nbulaj (approved):
LGTM actually, it looked terrifying at first look  :D

55728:
Thanks for taking the time to review! 👀✨ Yeah, the boot-order interaction is a bit mind-bending 😄

PR #315, opened 2026-06-17
Drop EOL Ruby 3.1, require Ruby >= 3.2

### LINKED ISSUE #314: CI fully red: i18n 1.15.0 drops Ruby 3.1 (Fiber[]), fail-fast cascades to all jobs
## Summary

As of today (2026-06-17), CI is red on `master` and on all open PRs. The specs themselves pass — the failure is environmental.

## Root cause

`i18n 1.15.0` was released today (2026-06-17 04:28 UTC). It uses the Fiber storage API `Fiber[:i18n_config]`, which only exists on **Ruby 3.2+**:

```
NoMethodError: undefined method `[]' for Fiber:Class
  current = Fiber[:i18n_config] || self.config = I18n::Config.new
  # i18n-1.15.0/lib/i18n.rb:58:in `config'
```

Our CI installs gems fresh (no committed lockfile for `gemfiles/*`), so it now resolves to `i18n 1.15.0`. The **Ruby 3.1** matrix jobs fail at load time (`require "active_record/railtie"`), and because the matrix has `fail-fast: true`, every other job is cancelled — so the whole run shows red even though Ruby 3.2/3.3/3.4 actually pass (e.g. `301 examples, 0 failures`, then `The operation was canceled.`).

Locally this doesn't reproduce because `Gemfile.lock` pins `i18n 1.14.8`.

## Context

- Ruby 3.1 reached **EOL on 2025-03-25**, so dropping it is reasonable.
- The gemspec has `required_ruby_version = '>= 3.1'`.

## Proposed fix

1. Replace `ruby: 3.1` with `ruby: 3.2` in the CI matrix (drop EOL, cover the real minimum), and update the related `exclude` entries.
2. Set `fail-fast: false` so one job's failure no longer masks the real status of the others.
3. Bump `required_ruby_version` in the gemspec from `'>= 3.1'` to `'>= 3.2'` to match.

Happy to open a PR for this. @nbulaj — does dropping Ruby 3.1 (and adding 3.2) sound good to you, or would you prefer to keep 3.1 by pinning `i18n < 1.15` for that row?

55728:
Closes #314

[i18n 1.15.0](https://github.com/ruby-i18n/i18n/releases/tag/v1.15.0) (released 2026-06-17) uses the `Fiber[]` storage API, which only exists on Ruby 3.2+. Since Ruby 3.1 reached EOL on 2025-03-25, this drops it from the CI matrix rather than pinning `i18n`.

**Changes:**

- Remove `ruby: 3.1` from the CI matrix and its two `exclude` entries (Rails 8.0 / doorkeeper master)
- Bump `required_ruby_version` in the gemspec from `'>= 3.1'` to `'>= 3.2'`
- CHANGELOG entry
- Bump `TargetRubyVersion` in the .rubocop.yml from `3.1'` to `3.2'`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,6 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         ruby:
-          - "3.1"
           - "3.2"
           - "3.3"
           - "3.4"
@@ -35,10 +34,6 @@ jobs:
           - gemfiles/rails_8.0.gemfile
           - gemfiles/doorkeeper_master.gemfile
         exclude:
-          - ruby: 3.1
-            gemfile: gemfiles/rails_8.0.gemfile
-          - ruby: 3.1
-            gemfile: gemfiles/doorkeeper_master.gemfile
           - ruby: head
             gemfile: gemfiles/rails_7.0.gemfile
     steps:
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,7 +4,7 @@ plugins:
   - rubocop-rails
   - rubocop-rspec
 AllCops:
-  TargetRubyVersion: 3.1
+  TargetRubyVersion: 3.2
   Exclude:
     # CI installs gems into ./vendor/bundle (ruby/setup-ruby bundler-cache),
     # which RuboCop would otherwise lint. Defining AllCops/Exclude overrides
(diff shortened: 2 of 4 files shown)


nbulaj (approved):
✅  LGTM, we can avoid Rubies which already EOL

55728:

PR #316, opened 2026-06-17
Set `fail-fast: false` in CI matrix

55728:
With `fail-fast: true`, a single failing matrix job cancels every other job — so the whole run shows red even when the other Ruby/Rails combinations pass. This made the `i18n 1.15.0` breakage (#314) look like a total CI failure when only the Ruby 3.1 row was actually broken.

Setting `fail-fast: false` lets each job report its real status independently.


nbulaj (approved):
Agree 👍

PR #325, opened 2026-07-04
Remove dead `Claims::AggregatedClaim` and `Claims::DistributedClaim` classes

55728:
`Claims::AggregatedClaim` and `Claims::DistributedClaim` were added in 2016 (25f2170) but have never been reachable:

- `lib/doorkeeper/openid_connect.rb` never requires their files
- `ClaimsBuilder` only constructs `NormalClaim`
- No spec, documentation, or application code references them
- The discovery endpoint accordingly advertises `claim_types_supported: ["normal"]` only

Aggregated and distributed claims (OIDC Core 1.0 §5.6) remain a possible future feature, but these empty `attr_accessor` shells carry no implementation to build on — whoever implements the feature will start from the spec, not from these.


nbulaj (approved):

55728:
Thank you @nbulaj 😊

PR #326, opened 2026-07-05
Harden the DCR spec against pre-existing `Application` rows

55728:
### Summary
The dynamic client registration spec asserted absolute DB state (`expect(Doorkeeper::Application.count).to eq(0|1)` in ten places, `Application.first` in six), so any `oauth_applications` row already present in `spec/dummy/db/test.sqlite3` (e.g. left behind by a `rake server` dashboard session against the same dummy app) made 12 of the 20 examples fail spuriously.

### Changes
Use `expect { post ... }.to change(Doorkeeper::Application, :count)` matchers and `Application.last`, so the examples assert what the request did rather than what the whole table contains.

### DIFF
--- a/spec/controllers/dynamic_client_registration_controller_spec.rb
+++ b/spec/controllers/dynamic_client_registration_controller_spec.rb
@@ -29,19 +29,21 @@ describe Doorkeeper::OpenidConnect::DynamicClientRegistrationController, type: :
   describe "#register" do
     context "when token_endpoint_auth_method is omitted" do
       it "defaults to client_secret_basic and creates a confidential client with a secret" do
-        post :register, params: {
-          client_name: "dummy_client",
-          redirect_uris: redirect_uris,
-          scope: "public",
-        }
+        expect do
+          post :register, params: {
+            client_name: "dummy_client",
+            redirect_uris: redirect_uris,
+            scope: "public",
+          }
+        end.to change(Doorkeeper::Application, :count).by(1)
 
         expect(response.status).to eq 201
-        expect(Doorkeeper::Application.count).to eq(1)
-
-        doorkeeper_application = Doorkeeper::Application.first
-        expect(doorkeeper_application.confidential).to be true
 
         body = JSON.parse(response.body)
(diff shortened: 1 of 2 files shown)


spec/controllers/dynamic_client_registration_controller_spec.rb
```diff
@@ -22,16 +22,17 @@
...
+            redirect_uris: redirect_uris,
+            scope: "public",
+          }
+        end.to change(Doorkeeper::Application, :count).by(1)
 
         expect(response.status).to eq 201
-        expect(Doorkeeper::Application.count).to eq(1)
 
-        doorkeeper_application = Doorkeeper::Application.first
+        doorkeeper_application = Doorkeeper::Application.last
```
nbulaj: I personally don't like idea of using last/first since it's very specific to the ordering and type of the OPK (which can be UUID v4 or whatever else). Which in turn can lead to flaky tests.

Ideally we have to find exactly processed application, say by name or whatever uniquely identifies it. Is it possible to do so? So instead of `.last` we will have `.find_by(name: "dummy_client")` for example.

55728: Thank you for the review 😊 Good point 👀✨ I've modified into `last` since all assertions pass even if unnecessary records remain on DB. Exactly, should be loaded by `find_by` for each assertion!

55728:
I believe switching back to `order(:id).last` wouldn't be appropriate here — the preceding review specifically asked to avoid `.last` / `.first` because row ordering depends on the PK type (e.g. UUID v4) and can lead to flaky tests.

The `change(...).by(1)` matchers already verify that a new row was created independently, so the `find_by(uid:)` lookup isn't circular in practice.

nbulaj:
I love the idea to lookup created record by the data from the response, that's how I mostly do on my project and which gives good guarantee that we found correct record :+1:

nbulaj (approved):
LGTM, nice work, thanks! 💪

55728:
Thank you for advising me to improve the examples further 👀✨

PR #329, opened 2026-07-07
Restore compatibility with the full `doorkeeper >= 5.5` range

### LINKED ISSUE #328: gemspec claims >= 5.5 but the suite fails on doorkeeper < 5.8
## Observation

The gemspec declares `doorkeeper >= 5.5`, but the code requires newer APIs. The actual breakage depends on the doorkeeper version:

| doorkeeper | `Orm::ActiveRecord::Mixins` (5.6.0+) | `pkce_code_challenge_methods` (5.8.0+) | Result |
|---|---|---|---|
| 5.5.x | ❌ | ❌ | **Load failure** — `NameError` at require time; the app won't boot at all |
| 5.6.0–5.7.x | ✅ | ❌ | **22 `NoMethodError` failures** (all from the PKCE call site) |
| 5.8.0+ | ✅ | ✅ | 303 examples, 0 failures ✅ |

The two call sites:

1. **`DiscoveryController#code_challenge_methods_supported`** calls `Doorkeeper.configuration.pkce_code_challenge_methods` (added in doorkeeper 5.8.0) — this is what causes the 22 runtime failures on 5.6–5.7.x
2. **`orm/active_record.rb`** prepends onto `Doorkeeper::Orm::ActiveRecord::Mixins::AccessGrant` at require time (the `Mixins` namespace was introduced in 5.6.0) — on 5.5.x this is a load-time `NameError`, not a test failure

So the effective floor is `>= 5.8`, and doorkeeper < 5.6 has been silently broken at the load level for some time.

## Reproduction

```sh
# Gemfile
gem "doorkeeper", "5.7.1"
gem "doorkeeper-openid_connect", "1.10.3"

$ bundle exec rspec
# => 22 failures (NoMethodError)
```

## Question for @nbulaj

Should the gemspec floor be raised to `>= 5.8`, or should the code maintain backward compatibility with 5.5?

If raising the floor is the right call, I'm happy to open a PR that bumps the gemspec and adds a `doorkeeper_5_8.gemfile` to CI so the floor stays tested going forward.

55728:
Fixes #328

## Summary

The gemspec declares `doorkeeper >= 5.5, < 6.0`, but in practice the gem only worked on Doorkeeper `5.8+`. This PR restores support for the whole declared range **without pinning the floor up to 5.8** — keeping the gem usable for large downstreams (GitLab, Mastodon, …) that track older Doorkeeper series.

The gap turned out to be much smaller than it looks: no polyfills, no vendored code — three narrow capability guards plus a couple of version-tolerant specs.

Verified locally:

| Doorkeeper | Result |
| --- | --- |
| 5.5.4 | 305 examples, **0 failures** |
| 5.6.7 | 305 examples, **0 failures** |
| 5.7.1 | 305 examples, **0 failures** |
| 5.9.3 | 307 examples, **0 failures** (no regression) |

RuboCop clean.

### 1. Doorkeeper 5.5.x — `NameError` at load time

`lib/doorkeeper/openid_connect/orm/active_record.rb` referenced `Doorkeeper::Orm::ActiveRecord::Mixins::AccessGrant` at require time, which raised `NameError` on 5.5.x and took the whole gem (and suite) down.

The key finding: **the mixin file has shipped since Doorkeeper 5.5.0**, and the `Doorkeeper::AccessGrant` model has always included it. What 5.5 lacks is only the *autoload registration* for the `Mixins` namespace — that was added in 5.6.0's rewrite of `orm/active_record.rb`. So this isn't a missing API; the constant simply isn't autoloadable yet.

Fix: require the file explicitly when the constant isn't already defined (a no-op on 5.6+, where it's autoloaded):

```ruby
unless defined?(Orm::ActiveRecord::Mixins::AccessGrant)
  require "doorkeeper/orm/active_record/mixins/access_grant"
end
```

The `included`-callback wiring introduced in v1.10.3 (#308) works unchanged on 5.5, since the mixin's `included do` block only runs at the host model's own load time.

### 2. Doorkeeper 5.6 / 5.7 — `pkce_code_challenge_methods` `NoMethodError`

`DiscoveryController#code_challenge_methods_supported` called `doorkeeper.pkce_code_challenge_methods`, which only exists since Doorkeeper 5.8.0 (doorkeeper-gem/doorkeeper@e39cd62c[^pkce]).

Fix: fall back to `%w[plain S256]` when the option is absent:

```ruby
return %w[plain S256] unless doorkeeper.respond_to?(:pkce_code_challenge_methods)
```

This is semantically exact, not a guess: `%w[plain S256]` is both the 5.8 default **and** what pre-5.8 Doorkeeper actually accepts whenever PKCE is available — it's also the value this gem itself hardcoded until it switched to delegation in `42d90dd`.

### 3. Specs — version-tolerant expectations

Three spec-only adjustments so the suite reflects each version's real behavior:

- **`pkce_code_challenge_methods` contexts** are skipped where the config option doesn't exist (`if: Doorkeeper::Config.method_defined?(...)`).
- **The `Scopes#allowed` fallback test** previously simulated old Doorkeeper by `undef_method(:allowed)`; on genuinely-old Doorkeeper there's nothing to undefine, so it now guards on `respond_to?(:allowed)`. (`Scopes#allowed` landed in 5.8.1, [doorkeeper-gem/doorkeeper@e2b8dabb][^allowed]; the production guard in `dynamic_registration_request.rb` was already correct — only the spec broke.)
- **Invalid pre-authorization specs** expect `200` before Doorkeeper 5.6.7 and `400` from 5.6.7 on, since error-status propagation landed in [doorkeeper-gem/doorkeeper@6cd5cca2][^status].

## CI

Adds `gemfiles/doorkeeper_5.{5,6,7}.gemfile` and three matrix rows (single Ruby each) so the declared `>= 5.5` floor stays honest going forward. The 5.8+ range is already covered by the existing `rails_*` and `doorkeeper_master` gemfiles.

[^pkce]: https://github.com/doorkeeper-gem/doorkeeper/commit/e39cd62c
[^allowed]: https://github.com/doorkeeper-gem/doorkeeper/commit/e2b8dabb
[^status]: https://github.com/doorkeeper-gem/doorkeeper/commit/6cd5cca2

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,18 @@ jobs:
         exclude:
           - ruby: head
             gemfile: gemfiles/rails_7.0.gemfile
+        # Oldest supported Doorkeeper series — one Ruby each is enough to
+        # keep the `>= 5.5` gemspec constraint honest.
+        include:
+          - ruby: "3.4"
+            os: ubuntu-latest
+            gemfile: gemfiles/doorkeeper_5.5.gemfile
+          - ruby: "3.4"
+            os: ubuntu-latest
+            gemfile: gemfiles/doorkeeper_5.6.gemfile
+          - ruby: "3.4"
+            os: ubuntu-latest
+            gemfile: gemfiles/doorkeeper_5.7.gemfile
     steps:
       - name: Repo checkout
         uses: actions/checkout@v3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased
 
+- [#329] Restore compatibility with the full `doorkeeper >= 5.5` range declared in the gemspec ([#328](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/issues/328)). Doorkeeper 5.5.x raised `NameError: uninitialized constant Doorkeeper::Orm::ActiveRecord::Mixins` at load time (the mixin file ships since 5.5.0 but is only autoloadable since 5.6.0 — it is now required explicitly when absent), and Doorkeeper 5.6/5.7 broke the discovery endpoint because `pkce_code_challenge_methods` only exists since 5.8.0 (the discovery response now falls back to `plain S256`, which is what those versions accept). CI now exercises the oldest supported Doorkeeper series via `gemfiles/doorkeeper_5.{5,6,7}.gemfile`
 - Please add here
 
 ## v1.10.4 (2026-07-07)
--- a/app/controllers/doorkeeper/openid_connect/discovery_controller.rb
+++ b/app/controllers/doorkeeper/openid_connect/discovery_controller.rb
@@ -83,6 +83,10 @@ module Doorkeeper
       def code_challenge_methods_supported(doorkeeper)
         return unless doorkeeper.access_grant_model.pkce_supported?
 
+        # Doorkeeper < 5.8 has no `pkce_code_challenge_methods` option and
+        # always accepts both methods whenever PKCE is available.
+        return %w[plain S256] unless doorkeeper.respond_to?(:pkce_code_challenge_methods)
+
         doorkeeper.pkce_code_challenge_methods
       end
 
--- a/gemfiles/doorkeeper_5.5.gemfile
+++ b/gemfiles/doorkeeper_5.5.gemfile
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+gem "doorkeeper", "~> 5.5.0"
+gem "rails", "~> 8.0.0"
+gem "rails-controller-testing"
+gem "sqlite3", "~> 2.3", platform: %i[ruby mswin mingw x64_mingw]
+
+gemspec path: "../"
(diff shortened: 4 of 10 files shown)


nbulaj (approved):
I love it, seems easy-peasy indeed 😍  Thanks!

55728:
Thanks for the quick review and merge! 😆🎉

nbulaj:
It was easier than I expected 😆

nbulaj:
Released as 1.10.5 to unblock users :eyes:

55728:
Thank you for the quick release! 👀✨🎉

iBrianCox:
Please remove me from this repo. I don’t know what this is.------------------------------On Jul 8, 2026, at 5:13 AM, Nikita Bulai ***@***.***> wrote:﻿
@nbulaj approved this pull request.

I love it, seems easy-peasy indeed 😍  Thanks!

nbulaj:
You just commented here yourself @iBrianCox , you're not a member of the repository/organization, not even sure how did you get here and why you received an email with the notification for this PR :no_good:  :thinking:

PR #332, opened 2026-07-15
Don't advertise refresh_token twice in grant_types_supported

55728:
## Summary

The discovery document's `grant_types_supported` (also used by Dynamic Client Registration to validate requested grant types) is built from the raw `grant_flows` configuration plus an unconditional `refresh_token` append when `use_refresh_token` is enabled. A configuration that both lists `refresh_token` in `grant_flows` explicitly *and* enables `use_refresh_token` therefore advertised the grant twice:

```ruby
Doorkeeper.configure do
  grant_flows %w[authorization_code refresh_token]
  use_refresh_token
end
# => "grant_types_supported": ["authorization_code", "refresh_token", "refresh_token"]
```

This is the same duplication Doorkeeper fixed for its own RFC 8414 metadata in doorkeeper-gem/doorkeeper#1847.

## Changes

Append the flow only when it is not already listed. The mixin keeps reading the raw `grant_flows` option (rather than delegating to Doorkeeper's calculated `enabled_grant_flows`) because extension-registered authorization flows such as `id_token` are part of the advertised list here, and the raw option is the only accessor shared by all supported Doorkeeper versions.

New unit specs cover the four `grant_flows` × `use_refresh_token` combinations and assert the configured array is not mutated.


nbulaj (approved):
👍  thanks!

PR #333, opened 2026-07-15
Add RFC 9207 companion support for doorkeeper#1849

55728:
## Summary

doorkeeper-gem/doorkeeper#1849 added RFC 9207 (OAuth 2.0 Authorization Server Issuer Identification) support to Doorkeeper: when `issuer` is configured (doorkeeper-gem/doorkeeper#1838), the `iss` parameter is emitted on authorization responses and `authorization_response_iss_parameter_supported` is advertised in the RFC 8414 metadata.

Doorkeeper can only do that for the response types and error paths it owns. The surfaces owned by this gem — the `id_token` / `id_token token` response types and the OIDC-specific error redirects (`login_required`, `consent_required`, ...) — were left out, so a server on the upcoming Doorkeeper release would advertise RFC 9207 support while OIDC responses lacked the parameter. This PR brings them in line.

## Changes

- **`IdTokenResponse` / `IdTokenTokenResponse`** merge `iss` into the fragment. The value is sourced from the response's own ID Token `iss` claim: RFC 9207 §2 requires the parameter to be *identical* to the `iss` claim of an ID Token returned from the authorization endpoint.
- **OIDC error redirects** forward Doorkeeper's issuer to the OAuth error response objects — the same wiring Doorkeeper uses for its own authorization errors (`pre_authorization.rb`). Doorkeeper gates the emission on redirectability, so non-redirectable errors never carry it; versions predating #1849 simply ignore the attribute.
- **Discovery document** advertises `authorization_response_iss_parameter_supported`, `true` exactly when Doorkeeper is configured with an issuer (mirroring Doorkeeper's own RFC 8414 document).
- **Boot-time consistency warning** when the OpenID Connect `issuer` and Doorkeeper's `issuer` are both statically configured with different values. RFC 9207-conforming clients compare the discovered issuer against the `iss` parameter and would reject every authorization response, so the misconfiguration is surfaced early. Callable issuers can't be compared statically and are skipped.

## Compatibility

Everything is gated on a new `Doorkeeper::OpenidConnect.doorkeeper_issuer` seam (`respond_to?`-guarded, shared with the #322 fallback), so there is **no behavior change** on released Doorkeeper versions or until an `issuer` is configured.

The boot-time check additionally sits behind a `doorkeeper_configured?` guard: reading Doorkeeper's config while it is not yet configured (initializer ordering is the host app's choice) raises `MissingConfiguration` on Doorkeeper 5.5 and eagerly builds a default configuration on newer versions. `Doorkeeper.configured?` only exists since 5.6 — reporting `false` on 5.5 is also correct there, since its config has no `issuer` to compare.


nbulaj (approved):
👍  LGTM

55728:
Thanks @nbulaj! 😊👍

PR #335, opened 2026-07-18
Add escape hatches for ID Token and User Info classes

### LINKED ISSUE #310: Allowing Custom ID Token/User Info Logic
Due to a series of previous Bad Decisions, I have run into a slightly strange case that changes the shape of the OIDC response based on the scopes selected, alongside other user configuration values. E.g., if the client chooses an `actor` scope, the response takes on a different shape than if they chose a `user` (`profile`) scope.

The existing DSL does not quite fit this particularly cleanly and would require one of two solutions: either passing `token` to all the procs to allow conditional logic, or allowing a complete override of the `id_token` and `user_info` classes to support significantly more in-depth customization while retaining the same "engine". Of these two, the latter seems to be much more interesting: it offers significantly more flexibility and power to evaluate based on system/business logic. The downside, of course, would that be this would basically make `claims` redundant if these values are set. `id_token` could also benefit from significantly expanded signing key resolution and other behaviors.

Would this be something desirable to have in this gem, or would such a divergence be considered out of scope for what this gem is hoping to do?

KazWolfe:
Fixes #310.

This pull request aims to allow developers to implement advanced logic for ID Token and User Info generation by providing escape hatches to override the default classes. This enables custom (context-based) claim management, as well as the potential to set or manage different signing keys for different contexts.

A few opinions were encoded in this change:

- We're replacing `IDToken` and `UserInfo` so that end-developers get full control over the OIDC deliverables.
  - I did consider replacing `ClaimsBuilder` with its own logic so that developers didn't have to duplicate code, but decided against it in favor of broadening possibilities. This does mean that devs using the escape hatch may need to implement things twice, but that explicit behavior might very well be intended.
  - I also do not force custom classes to inherit from `IDToken` or `UserInfo`, or really perform any validation on those classes. This is easy enough to add, but I deemed it unnecessary.
- This does *not* replace `claims`. By default, applications that use this gem will still go through the happy `claims` path. Developers leveraging the escape hatch may also choose if they want to continue using the ClaimsBuilder for their own purposes.


55728 (commented):
Thanks for putting this together, @KazWolfe 😊 — really nice to see #310 come to life! The approach is clean and the test coverage per call site is thorough 🎉

A couple of things I noticed while reading through:

**`id_token token` hybrid flow not covered**

`IdTokenTokenRequest#response` still instantiates `Doorkeeper::OpenidConnect::IdTokenToken` directly, so a custom `id_token_class` wouldn't take effect there. Since `IdTokenToken` inherits from `IdToken` and just adds `at_hash`, this could silently produce inconsistent claims in hybrid-flow responses. Would you be open to covering that path as well? Happy to brainstorm the best shape for it!

**Interface contract in docs may need `#issuer`**

The config comments and initializer template describe the minimum contract for a fully custom (non-subclass) ID Token as `#as_json`, `#as_jws_token`, and a matching initializer. However, since #333 (RFC 9207 support), `IdTokenResponse#body` calls `id_token.issuer` when Doorkeeper has an issuer configured — so a minimal non-subclass built from the docs alone would raise `NoMethodError` on `id_token` / `id_token token` authorization responses. Maybe worth adding `#issuer` to the documented contract, or leaning a bit harder on the subclass recommendation? 👀✨

KazWolfe:
I'll look more at dealing with `id_token_token` since that could be a bit painful and I want to avoid API breaks.

For the interface contract: it wouldn't be a problem to add at all, but I do want to ask if it would make sense to verify the interface or force things to be a subclass. I'd rather not use a subclass here, since the inheritance tree is not strictly necessary correct anymore.

Doorkeeper solves this problem by using [mixins](https://github.com/doorkeeper-gem/doorkeeper/blob/main/lib/doorkeeper/models/access_token_mixin.rb), but I'm not sure this library has enough to justify that.

KazWolfe:
Pushed a version that fixes both. I'm not particularly happy with either, but they should be serviceable.

- I moved `IdTokenToken`'s behavior to a `HybridIdTokenConcern` to be dynamically included when necessary. I don't think there's going to be much (any?) compelling reason to override this behavior, so having the library provide it should probably be fine. This, unfortunately, makes our proper `IdTokenToken` basically useless.
  - It's maintained for now (with some shim code to keep returning the same types just in case), but it should probably be slated for removal eventually.
- I updated documentation to reflect `#issuer`, and added validation by form of a config-time check. I'm *especially* not happy with this one, but it should be sufficient at least especially with the doc change.

55728 (approved):
Thanks for the quick turnaround on both points — the `HybridIdTokenConcern` extraction and the config-time validation are solid. This looks good to merge! 🎉

Thank you @KazWolfe! 😊

PR #336, opened 2026-07-19
Refactor `id_token_class` / `user_info_class` internals

55728:
## Summary

Reworks the internals behind the `id_token_class` / `user_info_class` configuration options: class resolution and validation now happen lazily at first use, the hybrid `id_token token` response is built through a single code path for default and custom classes alike, the implicit ivar contract in `HybridIdTokenConcern` becomes an explicit documented reader, and `Doorkeeper::OpenidConnect::IdTokenToken` is formally deprecated.

No changes to the configuration API — existing initializers keep working as-is.

### Constantize and validate the configured classes lazily

`Doorkeeper::OpenidConnect.configure` used to constantize and validate `id_token_class` / `user_info_class` while it ran. Since these options exist precisely so the host app can point at its own (reloadable) classes, that meant referencing app constants while initializers run — which breaks zeitwerk on Rails 7+, where reloadable constants must not be touched during boot. This is the same reason `open_id_request_model` already constantizes lazily.

Resolution and the required-methods check now happen at first use (`id_token_model` / `user_info_model`). The resolved class is deliberately **not** memoized, so code reloading in development never hands back a stale class; only the validation result is cached, keyed on the resolved class so a reloaded class is re-validated.

### Accept private implementations of the required methods

The required-methods check used `instance_methods`, which only reports public methods, so a class implementing e.g. `#issuer` as a private method was wrongly rejected. It now checks `method_defined?` / `private_method_defined?`.

### Single construction path for the hybrid `id_token token` response

`IdTokenTokenRequest` special-cased the shipped `IdTokenToken` class and only applied `HybridIdTokenConcern` to custom classes. Both now go through one path:

```ruby
id_token_model.new(auth.token, pre_auth.nonce).extend(HybridIdTokenConcern)
```

`extend` is a no-op when the concern is already in the ancestry (e.g. a configured subclass of `IdTokenToken`), so it cannot double-apply.

### Explicit `access_token` contract in `HybridIdTokenConcern`

The concern computed `at_hash` by reaching into the host object's `@access_token` ivar — an implicit contract a custom `id_token_class` could silently break. It now calls an `access_token` reader (private is fine); `IdToken` exposes it via `attr_reader`. The contract is documented in the concern, the `id_token_class` config option, and the initializer template.

### Deprecate `Doorkeeper::OpenidConnect::IdTokenToken`

With the unified hybrid path, the class only exists for API compatibility. It now emits a deprecation warning from its class body, and the eager `require` is replaced with an `autoload` so the warning fires only for code that actually references the constant — not on every boot. Migration path: `id_token_class` + `HybridIdTokenConcern`.

## Specs

- Configure-time validation specs rewritten for the lazy semantics (configure no longer raises; first use does).
- The `user_info_class` validation spec now has real teeth: `BasicObject` is the only class that escapes ActiveSupport's `Object#as_json`, so the failure branch is actually exercised.
- New coverage: private implementations of required methods, non-memoization across code reloads, the unified hybrid path (`IdToken` + concern, `at_hash` claim present), and the deprecation warning.
- Fixed a stale `AtHashConcern` reference in a spec description.

## CHANGELOG

Adds an entry for this rework and corrects the escape-hatch entry to point at [#335] (the PR that actually landed it; it was attributed to the superseded [#310]).


55728:
Reopening from an upstream branch so I can stack a follow-up PR on top. New PR: #337

PR #337, opened 2026-07-19
Refactor `id_token_class` / `user_info_class` internals

55728:
## Summary

Reworks the internals behind the `id_token_class` / `user_info_class` configuration options: class resolution and validation now happen lazily at first use, the hybrid `id_token token` response is built through a single code path for default and custom classes alike, the implicit ivar contract in `HybridIdTokenConcern` becomes an explicit documented reader, and `Doorkeeper::OpenidConnect::IdTokenToken` is formally deprecated.

No changes to the configuration API — existing initializers keep working as-is.

### Constantize and validate the configured classes lazily

`Doorkeeper::OpenidConnect.configure` used to constantize and validate `id_token_class` / `user_info_class` while it ran. Since these options exist precisely so the host app can point at its own (reloadable) classes, that meant referencing app constants while initializers run — which breaks zeitwerk on Rails 7+, where reloadable constants must not be touched during boot. This is the same reason `open_id_request_model` already constantizes lazily.

Resolution and the required-methods check now happen at first use (`id_token_model` / `user_info_model`). The resolved class is deliberately **not** memoized, so code reloading in development never hands back a stale class; only the validation result is cached, keyed on the resolved class so a reloaded class is re-validated.

### Accept private implementations of the required methods

The required-methods check used `instance_methods`, which only reports public methods, so a class implementing e.g. `#issuer` as a private method was wrongly rejected. It now checks `method_defined?` / `private_method_defined?`.

### Single construction path for the hybrid `id_token token` response

`IdTokenTokenRequest` special-cased the shipped `IdTokenToken` class and only applied `HybridIdTokenConcern` to custom classes. Both now go through one path:

```ruby
id_token_model.new(auth.token, pre_auth.nonce).extend(HybridIdTokenConcern)
```

`extend` is a no-op when the concern is already in the ancestry (e.g. a configured subclass of `IdTokenToken`), so it cannot double-apply.

### Explicit `access_token` contract in `HybridIdTokenConcern`

The concern computed `at_hash` by reaching into the host object's `@access_token` ivar — an implicit contract a custom `id_token_class` could silently break. It now calls an `access_token` reader (private is fine); `IdToken` exposes it via `attr_reader`. The contract is documented in the concern, the `id_token_class` config option, and the initializer template.

### Deprecate `Doorkeeper::OpenidConnect::IdTokenToken`

With the unified hybrid path, the class only exists for API compatibility. It now emits a deprecation warning from its class body, and the eager `require` is replaced with an `autoload` so the warning fires only for code that actually references the constant — not on every boot. Migration path: `id_token_class` + `HybridIdTokenConcern`.

## Specs

- Configure-time validation specs rewritten for the lazy semantics (configure no longer raises; first use does).
- The `user_info_class` validation spec now has real teeth: `BasicObject` is the only class that escapes ActiveSupport's `Object#as_json`, so the failure branch is actually exercised.
- New coverage: private implementations of required methods, non-memoization across code reloads, the unified hybrid path (`IdToken` + concern, `at_hash` claim present), and the deprecation warning.
- Fixed a stale `AtHashConcern` reference in a spec description.

## CHANGELOG

Adds an entry for this rework and corrects the escape-hatch entry to point at [#335] (the PR that actually landed it; it was attributed to the superseded [#310]).


lib/doorkeeper/oauth/id_token_token_request.rb
```diff
@@ -6,16 +6,12 @@ class IdTokenTokenRequest < IdTokenRequest
...
-        if token_model == Doorkeeper::OpenidConnect::IdToken
-          id_token_token = Doorkeeper::OpenidConnect::IdTokenToken.new(auth.token, pre_auth.nonce)
-        else
-          id_token_token = token_model.new(auth.token, pre_auth.nonce)
-                                      .extend(Doorkeeper::OpenidConnect::HybridIdTokenConcern)
-        end
+        # `extend` is a no-op when the concern is already in the ancestry
+        # (e.g. a configured subclass of the deprecated IdTokenToken), so the
+        # same path serves the default and any custom id_token_class alike.
+        id_token_token = Doorkeeper::OpenidConnect.configuration.id_token_model
```
KazWolfe, line 12: Gonna note that this is *strictly* an API break: you were returning an `IdTokenToken` before, but now you're not.

This probably doesn't matter much in Ruby, but if anyone is using `is_a?` or similar, this might start to fail.

55728: Good catch — you're right that this is technically a type-level API break.

In practice the risk is low (duck typing being the norm), but it's worth handling cleanly. Two options I see:

1. **CHANGELOG-only**: document the type change under a "Breaking" or "Migration" heading so anyone who does rely on `is_a?` knows to update. Minimal churn.
2. **Inheritance shim during deprecation**: make `IdTokenToken` inherit from `IdToken` (instead of just including the concern) so both `is_a?(IdTokenToken)` and `is_a?(IdToken)` hold true until #338 removes it. This buys full backward compat during the deprecation window, though it does change the class hierarchy.

I'd lean toward option 1 — it matches the deprecation warning already in place, and the likelihood of someone checking `is_a?(IdTokenToken)` in the wild is slim. But happy to go with option 2 if you or @nbulaj prefer the safer path.

Either way, I'll make sure the CHANGELOG entry calls this out explicitly.

nbulaj: Hey. Let's make it simpler so option 1 is quite good I think. I don't think we'll face any issues IRL, but changelog entry and https://github.com/doorkeeper-gem/doorkeeper-openid_connect/wiki/Migration-from-Old-Versions have to do their job

55728: Thanks @nbulaj! Option 1 it is.

I've pushed a CHANGELOG update (53da651) that calls out the type change as breaking with a link to the migration guide. I'll add a "1.11.0" section to the [wiki](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/wiki/Migration-from-Old-Versions) covering the lazy resolution change, the new `#access_token` reader requirement, the `IdTokenToken` deprecation, and the `is_a?` migration path.

spec/lib/id_token_token_spec.rb
```diff
@@ -1,6 +1,18 @@
...
+begin
+  original_stderr = $stderr
+  $stderr = StringIO.new
+  Doorkeeper::OpenidConnect::IdTokenToken
+ensure
+  $stderr = original_stderr
+end
 
 describe Doorkeeper::OpenidConnect::IdTokenToken do
   subject { described_class.new(access_token, nonce) }
```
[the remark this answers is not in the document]
55728: Thanks — I looked into moving this inside the RSpec lifecycle, but both variants trip cops this repo enforces at zero offenses: `before(:context)` hits `RSpec/BeforeAfterAll`, and doing it in a regular `before` hits `RSpec/ExpectOutput` (the cop only inspects `$stderr` assignment inside example groups, which is why the current top-level form is clean). Since we're actively working `.rubocop_todo.yml` down to nothing, adding a disable comment for this seems like the wrong trade.

On the risk itself: this is the only spec in the suite that references `Doorkeeper::OpenidConnect::IdTokenToken` (the deprecation assertion below uses `load`, so it's order-independent), and the redirection wraps exactly one statement with an `ensure` restore — so the window where stderr could be masked is the autoload of `id_token_token.rb` alone.

nbulaj (approved):

PR #341, opened 2026-07-20
Use Doorkeeper's `application_class` config for dynamic client registration

hzm-s:
The dynamic client registration controller called `Doorkeeper::Application.create!` directly instead of going through
`Doorkeeper.configuration.application_model`, ignoring any class configured via Doorkeeper's `application_class` option. This change fixes that.

See also: https://github.com/doorkeeper-gem/doorkeeper/blob/f84797d265f1c4e8de85baf05e54aa5edbcdea74/app/controllers/doorkeeper/applications_controller.rb#L30-L31


55728 (approved):
Thanks for the fix 😊 — the DCR controller was the only place that hard-coded `Doorkeeper::Application` instead of going through `Doorkeeper.configuration.application_model`. This brings it in line with Doorkeeper's own `ApplicationsController` 👍

Just two CI items to address before merge:

**RuboCop** — trailing comma needed after the last parameter of the multiline `stub_const` call (autocorrectable):

```
spec/controllers/dynamic_client_registration_controller_spec.rb:512:11
Style/TrailingCommaInArguments
```

**CHANGELOG Verifier** — needs an entry in `CHANGELOG.md`.

Both are minor — happy to merge once they're green 🟢

hzm-s:
Thanks for the review!

Fixed both - RuboCop offense and added a CHANGELOG entry.

55728:
Thanks for the quick fix! 👀✨

PR #342, opened 2026-07-20
ci: pin changelog-enforcer to v3.7.0 (node24 build)

55728:
## Why

Every run of the "Changelog verifier" job currently emits GitHub's Node 20 deprecation warning:

> Node 20 is being deprecated. This workflow is running with Node 24 by default. If you need to temporarily use Node 20, you can set the ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true environment variable. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

The workflow references the floating `v3` tag of `dangoslen/changelog-enforcer`, and that tag still resolves to a build whose `action.yml` declares `using: 'node20'`. The `v3.7.0` release is built with `using: 'node24'`, so pinning to it makes the warning go away (verified by comparing `action.yml` at both refs).

## What changed

Commit 1 — pin the action:

- `.github/workflows/changelog.yml`: `dangoslen/changelog-enforcer@v3` → `@v3.7.0`

Commit 2 — keep it (and the other actions) up to date going forward:

- `.github/dependabot.yml`: add a `github-actions` ecosystem entry, so future action releases (including the one that eventually drops node24) arrive as Dependabot PRs instead of deprecation warnings
- `.github/workflows/changelog.yml`: set `skipLabels: "Skip-Changelog,dependencies"`. Dependabot labels its PRs `dependencies` automatically (see e.g. #234), so its bumps pass the changelog check without an entry; `Skip-Changelog` (the action's default skip label) is kept as a manual escape hatch — note that label doesn't exist in the repo yet, create it if wanted
- `CHANGELOG.md`: entry added as `[#342]`

No behavior change to the enforcement of regular PRs — same major version of the action, just an exact pin to the release that carries the node24 runtime, plus the skip-label list.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,9 @@
 version: 2
 updates:
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: daily
 - package-ecosystem: bundler
   directory: "/"
   schedule:
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -13,3 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: dangoslen/changelog-enforcer@v3.7.0
+      with:
+        # Dependabot PRs are labeled "dependencies" automatically and don't
+        # need a changelog entry; "Skip-Changelog" keeps a manual escape hatch.
+        skipLabels: "Skip-Changelog,dependencies"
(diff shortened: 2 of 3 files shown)


nbulaj (approved):

nbulaj:
And I believe we will need the same one for the core lib :eyes:  Thanks!

55728:
Thank you for the review 😊 I will open the same PR on the core lib 👀✨

PR #346, opened 2026-07-21
Tests only - Add specs for uncovered branches reachable on current Doorkeeper

55728:
### Summary

This PR closes the test coverage gaps that correspond to real, reachable behavior on the current Doorkeeper release (5.9.x). Line coverage goes from 96.69% → 98.70%, branch coverage from 90.96% → 95.18%, with 15 new examples across 7 files.

### What's covered

- **Implicit OIDC flow E2E** (`AuthorizationsController`): approve `response_type=id_token` and `id_token token` (fragment redirect, nonce, aud, `at_hash` per OIDC Core §3.2.2.9) and deny via `#destroy` (`access_denied` in the fragment). Covers `IdToken`/`IdTokenToken` request strategies, `IdTokenRequest#deny`, and `IdTokenResponse#redirectable?`.
- **Callable `signing_algorithm`** configuration value.
- **Config validation**: `user_info_class` missing `as_json` → `InvalidConfiguration`; `jws_public_key` deprecation warning; `MissingConfiguration` when the gem is unconfigured.
- **`at_hash` digest selection**: HS384 (SHA-384) and fallback to SHA-256 for algorithms without a digest-length suffix (e.g. EdDSA).
- **`NormalClaim`** `#generator` / `#type` accessors.
- **Routes DSL block**: `controllers`, `skip_controllers`, and `as` options.
- **`AccessGrant` mixin** wiring deferred through an intermediate `ActiveSupport::Concern` (the non-Class branch of the `included` hook).

### What's intentionally left uncovered

The remaining uncovered lines/branches are old-Doorkeeper version compatibility paths (`id_token_request.rb:18`, `authorization/code.rb:18-20`, `password_access_token_request.rb:15-17`, `active_record.rb:55`, `openid_request.rb:22-23,33`) that are unreachable on Doorkeeper 5.9.x. Covering them would require stubbing `Gem.loaded_specs`, which is not representative of real behavior.

>[!NOTE]
>The `skip_controllers` routing spec keeps the discovery routes drawn rather than skipping all controllers, because `Journey::Routes#clear` does not invalidate the memoized recognition cache in actionpack 8.0.5 — only `add_route` does. An empty redraw leaves stale recognition behind. This has been reported upstream as [rails/rails#58170](https://github.com/rails/rails/pull/58170).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - [#310] Add escape hatches via `id_token_class` and `user_info_class` for advanced customization of the ID Token and UserInfo response objects.
 - [#341] Fix dynamic client registration to use `Doorkeeper.configuration.application_model` instead of `Doorkeeper::Application` directly, so that the `application_class` configuration is respected when registering clients.
 - [#342] CI: pin `dangoslen/changelog-enforcer` to `v3.7.0` so the changelog job runs on the Node 24 build of the action. The floating `v3` tag still points at a `node20` build, which triggers GitHub's [Node 20 deprecation warning](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) on every run. Dependabot now also watches GitHub Actions versions, and PRs labeled `dependencies` or `Skip-Changelog` are exempt from changelog enforcement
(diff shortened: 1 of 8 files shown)


nbulaj (approved):
More tests 😍

PR #347, opened 2026-07-22
Omit symmetric (oct) keys from the JWKS endpoint

55728:
A JWKS publishes *verification* keys, but an HMAC JWK **is** the shared secret itself: relying parties cannot use it to verify tokens the way they use a public key, so it does not belong in a public discovery document (RFC 7517).

`signing_keys_normalized` now drops `kty: "oct"` entries, so `/oauth/discovery/keys` returns an empty `keys` array for HMAC (HS256/HS384/HS512) configurations.

Note: on `jwt >= 2.5` the exported `oct` JWK already omits the secret material (`k`), so the previous behavior published a useless-but-harmless stub entry, not the secret. This change is spec hygiene — an empty `keys` array is the honest answer for a symmetric configuration — not a vulnerability fix.

Also folds in a small tidy-up: `signing_algorithm` now resolves callables through the existing `unwrap_callable` helper instead of an inline `respond_to?(:call)` branch (keeps the module within RuboCop's `Metrics/ModuleLength` default after the rebase).


nbulaj (approved):
LGTM  ✅

PR #348, opened 2026-07-22
Harden the OIDC authorization endpoint against malformed requests

55728:
Three independent robustness fixes for the authorization endpoint, each with a regression spec that fails on the old code:

### 1. Non-scalar `max_age` caused a 500

A request like `GET /oauth/authorize?...&max_age[]=1` arrives as an `Array` (or `ActionController::Parameters` for a hash), which does not respond to `#to_i`, so `handle_oidc_max_age_param!` raised `NoMethodError` and returned a 500.

`max_age` is a single string value per OIDC Core §3.1.2.1. Parsing now goes through `oidc_max_age_seconds`, which accepts only `String`/`Integer` and returns `nil` otherwise, so a malformed `max_age` is treated as absent.

### 2. `prompt` handling mutated the shared `request.query_parameters`

`return_without_oidc_prompt_param` stripped the `prompt` value out of `request.query_parameters` in place via `.tap`. That hash is memoized and shared for the lifetime of the request, so the mutation corrupted the parameters seen by everything after it — including any subsequent prompt value processed in the same loop (`prompt=login select_account`). The return URL is now built from a `dup`.

### 3. Internal configuration errors leaked to the client as OAuth errors

`authenticate_resource_owner!` rescued `Errors::OpenidConnectError` — the base class — so an internal `Errors::InvalidConfiguration` (e.g. from an unconfigured `reauthenticate_resource_owner` / `select_account_for_resource_owner` callback) was caught and redirected to the client as a bogus `error=invalid_configuration` authorization response. A server misconfiguration is not a protocol error the client can act on.

This introduces an intermediate `Errors::AuthorizationError` grouping the redirectable OAuth/OIDC protocol errors (`InvalidRequest`, `LoginRequired`, `ConsentRequired`, `InteractionRequired`, `AccountSelectionRequired`), and rescues only that. Internal errors now surface as 500s.

>[!NOTE]
> ### Compatibility note
> the protocol error classes remain `OpenidConnectError` subclasses, so existing `rescue OpenidConnectError` handlers in host applications are unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 - [#342] CI: pin `dangoslen/changelog-enforcer` to `v3.7.0` so the changelog job runs on the Node 24 build of the action. The floating `v3` tag still points at a `node20` build, which triggers GitHub's [Node 20 deprecation warning](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) on every run. Dependabot now also watches GitHub Actions versions, and PRs labeled `dependencies` or `Skip-Changelog` are exempt from changelog enforcement
 - [#346] Add specs for previously uncovered branches that are reachable on current Doorkeeper
 - [#347] Omit symmetric (`oct`) keys from the JWKS endpoint when signing with an HMAC algorithm — an HMAC JWK is the shared secret, not a verification key (RFC 7517)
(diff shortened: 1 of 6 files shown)


nbulaj (approved):
Love this `AuthorizationError` change 👍

PR #349, opened 2026-07-22
Fix claims pipeline: response symbolization, owner-less/app-less tokens, duplicate lookup

55728:
Five fixes in the claims/ID-token pipeline, each with a regression spec that fails on the old code:

### 1. String `response:` values silently dropped claims

`ClaimsBuilder` dispatches claims by comparing against the symbol tokens `:id_token` / `:user_info`, but a claim declared with string responses —

```ruby
claim :foo, response: ["id_token"] do |owner|
  ...
end
```

— stored the strings as-is, so `"id_token" == :id_token` was always false and the claim was silently dropped from **every** response. `Claim#initialize` now symbolizes the configured response values, making string- and symbol-configured claims behave identically.

### 2. Owner-less access token caused a 500 at the token endpoint

A `client_credentials` grant that carries the `openid` scope has no resource owner. `TokenResponse#body` still built an ID Token for it; resolving the nil owner then blew up in the `subject` block / claim generators with `NoMethodError` (500).

An ID Token's `sub` identifies an end user, and there is none here — so the token response now simply omits `id_token` for a token with no `resource_owner_id`. (Rebased on top of #310: the on-the-fly token is built through `configuration.id_token_model`.)

### 3. Owner-less access token caused a 500 at the userinfo endpoint, too

The same class of token — `client_credentials` with the `openid` scope, or a token whose owner was **deleted after issuance** — also crashed the userinfo endpoint: the claim generators dereferenced the nil owner (`NoMethodError` → 500).

The UserInfo response is a statement about the End-User, so a token that no longer resolves to one is not valid at this endpoint: it now answers with RFC 6750's `401 invalid_token` (via `Doorkeeper::OAuth::InvalidTokenResponse`, including the `WWW-Authenticate` header). `UserInfo#resource_owner` is now public so the controller can check resolution without adding a second lookup (it shares the memoized value with claim generation).

### 4. Application-less access token caused a 500 via `MissingRequiredClaim(:aud)`

`aud` is a REQUIRED claim sourced from the application's uid. A token with no application (e.g. a password grant with client authentication skipped) made ID Token serialization raise `MissingRequiredClaim` (500). The on-the-fly builder now skips the ID Token for application-less tokens, the same way as the owner-less case.

### 5. Resource owner was resolved twice per request

`UserInfo` and `IdToken` resolve the resource owner via `resource_owner_from_access_token`, and then `ClaimsBuilder.generate` looked it up a *second* time. `ClaimsBuilder.generate` now accepts the already-resolved owner from the caller — one DB lookup per response instead of two. A private sentinel distinguishes "not passed" from a legitimately-nil owner, so a nil owner does not trigger a re-lookup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
 - [#349] Claims pipeline fixes:
   - Dispatch claims whose `response:` option is configured with strings (e.g. `response: ["id_token"]`) instead of silently dropping them
   - Omit `id_token` from the token response when the access token has no resource owner (e.g. a `client_credentials` grant with the `openid` scope) instead of returning a 500
+  - Likewise omit `id_token` when the access token has no application, since the REQUIRED `aud` claim cannot be sourced (previously a 500 via `MissingRequiredClaim`)
+  - Answer userinfo requests whose token no longer resolves to a resource owner (`client_credentials` tokens with the `openid` scope, owners deleted after issuance) with RFC 6750's `401 invalid_token` instead of a 500
   - Resolve the access token's resource owner once per UserInfo / ID Token response instead of twice
 - [#350] Discovery / Dynamic Client Registration spec compliance fixes:
   - Advertise and accept the standard `implicit` grant type instead of doorkeeper's internal `implicit_oidc` name (RFC 7591 §2)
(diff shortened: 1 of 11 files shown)


nbulaj (approved):
Nice work! 👍

55728:
Thank you @nbulaj! 😊🙌

PR #350, opened 2026-07-22
Align discovery and DCR responses with RFC 7591 / RFC 7033

55728:
## Summary

Spec-compliance fixes for the discovery document, Dynamic Client Registration, and WebFinger:

### 1. `grant_types_supported` leaked the internal `implicit_oidc` flow name

This gem registers doorkeeper's `implicit_oidc` grant flow internally for the `id_token` / `id_token token` response types, and that internal name leaked verbatim into the discovery document and DCR responses. It is now mapped to the standard `implicit` grant type; DCR registration requesting `implicit` is accepted, and the internal `implicit_oidc` name is rejected as `invalid_client_metadata`.

### 2. DCR defaults claimed every server-supported type

When `response_types` / `grant_types` are omitted from a registration request, the response claimed the client registered for everything the server supports. RFC 7591 §2 specifies the defaults: `response_types` → `["code"]`, `grant_types` → `["authorization_code"]`.

### 3. DCR used an invented error code

Registration failures returned `invalid_client_params`, which appears in no RFC. RFC 7591 §3.2.2 defines the registration error codes; the endpoint now returns `invalid_redirect_uri` for redirect URI validation failures and `invalid_client_metadata` for everything else.

### 4. WebFinger response headers

The WebFinger endpoint now serves `application/jrd+json` and `Access-Control-Allow-Origin: *` per RFC 7033 §5 / §10.2. (A missing `resource` parameter already maps to 400 via `ParameterMissing`.)

### 5. Custom application model support

Clients are created through `Doorkeeper.configuration.application_model` instead of the hardcoded `Doorkeeper::Application`, so installations with a custom application model can use dynamic registration. (Master already made this change for the `create!` call in #341; this branch's remaining diff aligns the rest.)

>[!NOTE]
> Item 3 changes an error code that clients may have been matching on (`invalid_client_params` → RFC codes). It was never a registered code, but calling it out as the one behavior change with a (small) compatibility surface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,12 @@
   - Treat a malformed, non-scalar `max_age` parameter (e.g. `max_age[]=1`) as absent instead of returning a 500 (OIDC Core §3.1.2.1)
   - Build the `prompt=login` / `prompt=select_account` return URL from a copy of `request.query_parameters` instead of mutating the shared, memoized hash
   - Redirect only OAuth/OIDC protocol errors (grouped under the new `Errors::AuthorizationError`) to the client; internal errors like `Errors::InvalidConfiguration` now propagate as a 500 instead of leaking as a spurious authorization error
+- [#350] Discovery / Dynamic Client Registration spec compliance fixes:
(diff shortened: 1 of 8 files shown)


nbulaj (approved):
👍  thanks

55728:
Thank you @nbulaj 😊

PR #351, opened 2026-07-22
Align OIDC token/authorization responses with doorkeeper core

55728:
## Summary

Three places where this gem's response classes diverge from doorkeeper core's, each with a regression spec that fails on the old code:

### 1. `id_token token` returned the stored token, not the plaintext

`IdTokenTokenResponse#body` returned `auth.token.token` for the implicit flow's `access_token`. With a hashing token-secret strategy (`hash_token_secrets`), the stored `token` attribute is the **hash**, so the client received an unusable access token. It now uses `plaintext_token`, matching core `CodeResponse` / `TokenResponse`.

### 2. Missing `issued_token` broke `after_successful_authorization` hooks

For the `id_token` and `id_token token` response types, `strategy.authorize` returns this gem's response objects, and doorkeeper's `Hooks::Context#issued_token` calls `auth.issued_token` on them. Core's `CodeResponse` / `TokenResponse` implement it; ours didn't, so a host application's `after_successful_authorization` hook reading `context.issued_token` raised `NoMethodError`. Both response classes now return the issued access token.

### 3. Auth-code flow fired the response hook before attaching `id_token`

The authorization-code flow attached the `id_token` to the response *after* calling `super` (which fires `after_successful_strategy_response`), while the password flow attaches it *before*. A consumer hook therefore saw the complete response (including `id_token`) in the password flow but an `id_token`-less one in the auth-code flow. The auth-code flow now attaches before `super`, matching the password flow. (Rebased on top of #310: the ID Token is built through `configuration.id_token_model`.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@
   - Treat a malformed, non-scalar `max_age` parameter (e.g. `max_age[]=1`) as absent instead of returning a 500 (OIDC Core §3.1.2.1)
   - Build the `prompt=login` / `prompt=select_account` return URL from a copy of `request.query_parameters` instead of mutating the shared, memoized hash
   - Redirect only OAuth/OIDC protocol errors (grouped under the new `Errors::AuthorizationError`) to the client; internal errors like `Errors::InvalidConfiguration` now propagate as a 500 instead of leaking as a spurious authorization error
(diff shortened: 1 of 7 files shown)


nbulaj (approved):
LGTM 👍

55728:
Thanks for the review! 👀✨

PR #354, opened 2026-07-24
Remove stale TODO/FIXME markers and the dead compat code behind them

55728:
Prunes four stale TODO/FIXME markers and the dead code they guarded, one commit per marker so each removal carries its own rationale in the commit message. No behavior change.

## What was removed (and why it's safe)

Both compat shims below were introduced together in 430413d ("Add support for Doorkeeper 5.4", 2020) and self-declared their removal condition — both conditions have long been met by the `doorkeeper >= 5.5` gemspec requirement.

- **`current_resource_owner` override in `Helpers::Controller`** — marked "FIXME: remove after Doorkeeper will merge it". Doorkeeper merged the identical `defined?` memoization in 5.5.0 ([doorkeeper#1410](https://github.com/doorkeeper-gem/doorkeeper/pull/1410)), which is exactly the gemspec floor, so upstream now does the same thing on every supported version.
- **Doorkeeper < 5.4 `issue_token` fallback paths** — the `method_defined?(:issue_token!)` guard and its `else` branch in `OAuth::Authorization::Code`, plus the matching `respond_to?(:issue_token!)` checks in `IdTokenRequest` and the two id_token response specs. `issue_token!` exists on every installable version, so the legacy side was unreachable. The public `issue_token` alias is kept, so the gem's API is unchanged.
- **"support id_token response type" TODO in the discovery controller** — `id_token` / `id_token token` flows have been registered by this gem for years, and `response_types_supported` is built dynamically from `doorkeeper.authorization_response_types`.

## What was deliberately NOT removed

The `@_response_body = nil` line in `clear_oidc_response` was marked "FIXME: workaround for Rails 5", but [rails/rails#25106](https://github.com/rails/rails/issues/25106) is still present in Rails 8: `response_body = nil` only resets the response object's body and leaves `@_response_body` assigned, which would keep `performed?` returning true. The FIXME is replaced with a comment stating the constraint so the line doesn't read as removable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
   - Return the plaintext access token from the `id_token token` implicit response, so it stays usable when `hash_token_secrets` is enabled
   - Define `#issued_token` on `IdTokenResponse` / `IdTokenTokenResponse`, so `after_successful_authorization` hooks can read `context.issued_token` without raising
   - Attach the ID token before the `after_successful_strategy_response` hook fires in the authorization-code flow, matching the password flow
+- [#354] Remove stale TODO/FIXME markers and the Doorkeeper < 5.5 bridging code behind them, made obsolete by the `doorkeeper >= 5.5` requirement. No behavior change; see the individual commits for details
 - Add entry here
 
 ## v1.10.5 (2026-07-09)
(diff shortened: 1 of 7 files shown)


nbulaj (approved):
[Looks good](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/blob/master/doorkeeper-openid_connect.gemspec#L39) 👍 

Appreciate that 🙇

PR #355, opened 2026-07-25
Follow-up to #351: compute at_hash from the plaintext access token

55728:
Completes the response-contract alignment started in #351.

#351 switched the hybrid `id_token token` response body to return `plaintext_token`, so the access token stays usable when Doorkeeper's `hash_token_secrets` strategy is enabled. However, the `at_hash` claim inside the ID Token was still computed over the *stored* token value — which, under a hashing strategy, is a SHA-256 digest rather than the value the client receives.

Per OIDC Core 1.0 §3.2.2.9, `at_hash` must be derived from the octets of the access token as delivered to the client, so a conforming RP hashing the token it received would never match and would reject an otherwise valid ID Token. This only bites with `hash_token_secrets` enabled (non-default) in the hybrid flow; with the default plain strategy `plaintext_token == token`, so behavior there is unchanged.

Changes:

- `HybridIdTokenConcern#at_hash` now digests `@access_token.plaintext_token`, mirroring the response-body change from #351.
- Added a spec simulating a hashing token-secret strategy (stubbing the `plaintext_token` / `token` seam, same pattern as the #351 specs) that asserts `at_hash` is derived from the plaintext value.

> [!NOTE]
> 
> No `plaintext_token == nil` path is reachable here: the ID Token is built at issuance time (raw token in memory), and Doorkeeper itself force-disables `reuse_access_token` when the token secret strategy cannot restore secrets. Doorkeeper core's own `TokenResponse` calls `plaintext_token` unconditionally under the same assumptions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
   - Define `#issued_token` on `IdTokenResponse` / `IdTokenTokenResponse`, so `after_successful_authorization` hooks can read `context.issued_token` without raising
   - Attach the ID token before the `after_successful_strategy_response` hook fires in the authorization-code flow, matching the password flow
 - [#354] Remove stale TODO/FIXME markers and the Doorkeeper < 5.5 bridging code behind them, made obsolete by the `doorkeeper >= 5.5` requirement. No behavior change; see the individual commits for details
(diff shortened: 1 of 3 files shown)


nbulaj (approved):
👍

PR #356, opened 2026-07-29
Probe Doorkeeper's metadata endpoint without resolving `Object`

55728:
## Summary

The Doorkeeper 6.0 capability check added in #353 asks `Doorkeeper::OAuth.const_defined?(:MetadataResponse)` with the default `inherit: true`. That lookup continues into `Object`, so **any host application that happens to define a top-level `MetadataResponse` constant sends Doorkeeper 5.x down the 6.0 branch — and the application cannot boot**:

```
NameError: uninitialized constant Doorkeeper::MetadataController
# .../lib/doorkeeper/openid_connect/engine.rb:17:in 'block in <class:Engine>'
```

`#353` has not shipped in a release yet, so no released version is affected.

## Root cause

`const_defined?` and the `::` scope operator disagree about `Object`:

```ruby
$ ruby -e 'module A; module B; end; end; class MetadataResponse; end
          p A::B.const_defined?(:MetadataResponse)         # => true
          p A::B.const_defined?(:MetadataResponse, false)  # => false
          A::B::MetadataResponse'                          # => NameError
```

So the probe answers "yes, Doorkeeper 6.0" while the very next line — which reaches for `Doorkeeper::OAuth::MetadataResponse` and `Doorkeeper::MetadataController` through `::` — raises. It is enough for the host app's constant to be *registered* rather than loaded: `const_defined?` is true for a pending Zeitwerk autoload, so an ordinary `app/**/metadata_response.rb` is sufficient.

`MetadataResponse` is a plausible name in any application that serves metadata of its own, which is what makes this worth fixing rather than filing as a curiosity.


nbulaj (approved):
just a general observation: we have to move to `autoload` instead of `require` at some point I think for the lib/doorkeeper/openid_connect.rb 👀

55728:
Good observation, thanks! 👀✨ That sounds like a nice improvement to explore 👍

55728:
Thanks for the lightning-fast review! ⚡

PR #357, opened 2026-07-29
Test: Cover the conditional members of the enriched RFC 8414 document

55728:
`registration_endpoint` and `end_session_endpoint` were only ever asserted in their omission branches: the registration example requires the route to be absent, and the dummy app never configures an end-session endpoint. Deleting either line from `MetadataExtension#oidc_metadata` therefore left the whole suite green — neither member was pinned to anything.

This adds the positive branch for both:

- The registration example mirrors the existing omission example — it enables `dynamic_client_registration`, redraws the routes and forces Rails' lazy route loading to resolve now, so the premise that the URL helper exists is asserted rather than assumed.
- The end-session example additionally pins that the configured block is evaluated in the controller's context (`instance_exec`), matching the discovery document's own example.

Spec-only, no behavior change. Verified against Doorkeeper 6.0 (the file is gated on `Doorkeeper::OAuth::MetadataResponse`, so it is exercised by the `doorkeeper_6.0` gemfile leg): full suite green, and removing either member's line from `MetadataExtension#oidc_metadata` now fails exactly its example.

### DIFF
--- a/spec/controllers/doorkeeper/metadata_controller_spec.rb
+++ b/spec/controllers/doorkeeper/metadata_controller_spec.rb
@@ -9,6 +9,16 @@ require "rails_helper"
 # referencing Doorkeeper::MetadataController on a version that lacks it.
 if Doorkeeper::OpenidConnect.doorkeeper_metadata_endpoint?
   describe Doorkeeper::MetadataController, type: :controller do
+    # Restore the routes mutated by the registration_endpoint examples so no
+    # drawn route leaks into the rest of the suite. The configuration has to
+    # be restored first: this hook runs before the config-level reset in
+    # rails_helper, and reloading the routes while dynamic_client_registration
+    # is still enabled would redraw the very route being cleaned up.
+    after do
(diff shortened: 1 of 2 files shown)


spec/controllers/doorkeeper/metadata_controller_spec.rb
```diff
@@ -75,6 +75,47 @@
         expect(data).not_to have_key("jwks_uri")
       end
 
+      it "advertises the registration_endpoint when dynamic client registration is enabled and its route is drawn" do
```
nbulaj, line 88: Can we add a suite-scoped `after {}` hook with `Rails.application.reload_routes!` in order not to mutate state between tests?

55728: Good catch, added 😊

nbulaj: Thanks!

nbulaj (approved):
✅

PR #359, opened 2026-07-30
Add eager-load CI job for autoload migration safety

55728:
## Why

A broken `require` fails at boot — CI catches it. A broken `autoload` fails at the first reference to the constant, which in a host app means a `NameError` on some request after release. This job restores the boot-time safety net before the migration PRs land.

## What

**`bundle exec rake eager_load`** boots `spec/dummy` with `config.eager_load = true`, forces every pending `autoload` this gem owns, and fails when any `lib/` file ends up loaded by nothing. Covers wrong paths, wrong constant names, and dropped requires with no replacement.

**CI** (`.github/workflows/eager_load.yml`) gates the job behind a diff check — it only runs when `require`/`autoload` lines change, or files are added/removed/renamed under `lib/` or `app/`. The `eager-load-check` label requests it on any PR. Matrix: the 5 Doorkeeper series × 2 Rails (7 cells).

**Dummy cleanup.** Removes four unused `rails new` stubs (`ApplicationCable::*`, `ApplicationJob`, `ApplicationMailer`) that reference railties the test app doesn't load — they `NameError` under eager loading.


nbulaj:
I have a feeling that we're trying to solve some core design issue, but do it wrongly :eyes: Code looks like some very-very tricky hacking which inspects some internals and so on. Which exactly issue we're trying to solve? Host app with doorkeeper openid connect can't resolve some require or whatever? Shouldn't it be just fixed in the library?

55728:
Fair point, and the ordering is my fault — this is the wrong end of the stack to read the "why" from.

**Nothing is broken today.** On master `lib/doorkeeper/openid_connect.rb` has 33 `require`s and zero `autoload`s, and that works fine. This PR is the safety net for the migration you already approved in #360 — not a fix for a host-app issue.

But here's the thing: **#360 doesn't actually depend on this PR.** It touches none of the files introduced here; I only stacked it underneath on principle ("net before trapeze"). So I'll rebase #360 onto master and unstack it.

As for this PR — your proportionality instinct is right. The constant sweeper + 7-cell CI matrix is more machinery than the migration warrants. I'm closing this one. 🙏

PR #360, opened 2026-07-31
Convert OpenidConnect-namespaced requires to autoload (phase 1)

55728:
## What moves

Thirteen files that define exactly one constant and do nothing else → `autoload`, the way `lib/doorkeeper.rb` wires Doorkeeper's own:

`ClaimsBuilder`, `Claims::Claim`, `Claims::NormalClaim`, `DiscoveryHelpersMixin`, `Errors`, `GrantTypesSupportedMixin`, `HybridIdTokenConcern`, `IdToken`, `IdTokenToken`, `TokenEndpointAuthMethodsSupportedMixin`, `UserInfo`, `OAuth::DynamicRegistrationRequest`, and the five `version.rb` constants.

## What stays a require

Each annotated in source with the reason — in short:

- **`config.rb`** — also defines `.configure` / `.configuration` on the module; autoloading `Config` wouldn't trigger those.
- **`engine.rb`** — `Rails::Engine` subclass must exist by the time the gem finishes loading.
- **`helpers/controller.rb`** and the five **`oauth/*`** files — end with a `prepend` onto Doorkeeper's classes; nothing names the module afterwards, so no reference would trigger the autoload.

## Load-time side-effects spec

Autoloading `helpers/controller.rb` was the mistake that motivated this: the eager-load check stays green (forcing the autoload *is* the prepend), but 20 controller specs die with no hint of the cause. `spec/lib/load_side_effects_spec.rb` now asserts every prepend is in place after a plain boot via ancestor name comparison — complementary to `rake eager_load`, not redundant with it.

## Verification

- `rake spec` + `rake eager_load` — green on all 7 gemfiles
- `rubocop` — no offenses


lib/doorkeeper/openid_connect.rb
```diff
@@ -4,6 +4,54 @@
...
+    # references the deprecated constant instead of every boot.
+    autoload :IdTokenToken, "doorkeeper/openid_connect/id_token_token"
+    autoload :TokenEndpointAuthMethodsSupportedMixin,
+             "doorkeeper/openid_connect/token_endpoint_auth_methods_supported_mixin"
+    autoload :UserInfo, "doorkeeper/openid_connect/user_info"
+
+    # `version.rb` defines five sibling constants rather than one namespace to
+    # hold them, and an autoload fires only for the constant it names, so
+    # reading `MAJOR` must not depend on someone having touched `VERSION` first.
+    autoload :MAJOR, "doorkeeper/openid_connect/version"
```
nbulaj: This looks.. strange at least 🙂  Do we really need to autoload them like that? I remember Doorkeeper core [loads just VERSION module](https://github.com/doorkeeper-gem/doorkeeper/blob/main/lib/doorkeeper/version.rb), maybe we can simplify here as well?

55728: Good call — simplified to a single `autoload :VERSION`. The sibling constants (`MAJOR`, `MINOR`, `TINY`, `PRE`) are all defined the moment `version.rb` loads, and nothing references them standalone; their only other reader, the gemspec, requires the file directly.

I didn't go all the way to core's `module VERSION` structure: ours exposes the version string itself (`spec.version = Doorkeeper::OpenidConnect::VERSION`), so wrapping it in a namespace module would be a type-level breaking change — better suited to a dedicated follow-up 👀✨

nbulaj (approved):
LGTM ✅

55728:
Thanks for the review @nbulaj 👀✨

nbulaj:
Hope nobody verifies the version with `if Doorkeeper::OpenidConnect::MAJOR > 1` :eyes: 

```
lib/doorkeeper/openid_connect/version.rb defines MAJOR, MINOR, TINY, PRE, and VERSION, but only VERSION is registered for autoload. Referencing Doorkeeper::OpenidConnect::MAJOR (or the others) before VERSION would now raise NameError, which is a backwards-incompatible change from the previous unconditional require.
```

But I'd say we can ignore :slightly_smiling_face:  Especially for the 2.0 release which is planned

55728:
Fair 😊— and that's exactly the trade I made when I collapsed them: the four siblings had their own autoload precisely because they're flat constants, so VERSION wasn't the only door into the file. Agreed it's not worth the lines back, though. Nothing in the gem touches them, and everyone else nests these under a module (Rails::VERSION::MAJOR, Doorkeeper::VERSION::MAJOR), so the idiom people actually reach for never worked here anyway.

Which tempts me to adopt that shape in 2.0 — module VERSION with STRING: one autoload, all five constants, and it needs the major bump anyway since VERSION goes String → Module 🎉

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@
...
 - [#357] Add specs for the conditional `registration_endpoint` and `end_session_endpoint` members of the enriched RFC 8414 document, which were previously only asserted in their omission branches
+- [#360] Load the OpenID Connect-namespaced constants with `autoload` instead of `require`, the way `lib/doorkeeper.rb` loads Doorkeeper's own. Thirteen files — the claims classes, `Errors`, `IdToken`, `UserInfo`, `ClaimsBuilder`, `HybridIdTokenConcern`, the three discovery mixins, the version constants and the dynamic client registration request — are now loaded on first reference rather than on every boot. No public constant changes name or moves. Files that only exist for a load-time side effect (`config.rb`, `engine.rb`, `helpers/controller.rb` and the five `oauth/*` files that prepend onto Doorkeeper's classes) keep their `require`, with a comment on each explaining why; a new spec asserts every one of those prepends is in place after a plain boot, so a later conversion cannot silently drop one
```
nbulaj: Let's try to make changelog entries smaller since it can be hard for readers to quickly check the changelog. I believe we even did some PR with the same purposes

nbulaj (approved):

CHANGELOG.md
55728: Rewritten to one line:

```
- [#360] Load the OpenID Connect constants with `autoload` instead of `require`. No public constant changes name or moves
```

55728:

PR #362, opened 2026-07-31
Migrate Doorkeeper-extending requires to autoload (phase 2)

55728:
## Key insight

"Extends Doorkeeper's namespace" turned out to be two things:

- **Additive** — `Doorkeeper::OAuth::IdTokenRequest`, `IdTokenResponse`, `IdTokenTokenRequest`, `IdTokenTokenResponse`, and (on Doorkeeper ≥ 6.0) `MetadataResponse` only add constants. Safe to autoload.
- **Side-effecting** — files ending with `SomeDoorkeeperClass.prepend OurModule` are loaded *for* that statement. Nothing names the module → no autoload trigger → the prepend silently never happens. Must stay `require`.

Only the additive ones move. The side-effecting ones stay with the reason annotated in source.

## What stays a require (and why)

- **`request/id_token.rb`** / **`request/id_token_token.rb`** — `GrantFlow.register` stores the strategy class, not a name; autoload would fire immediately and defer nothing.
- **`orm/active_record.rb`** — prepends onto Doorkeeper's access grant mixin at load time.
- **`rails/routes.rb`** — loaded by the engine initializer on every boot anyway.

## Result

| | before | after |
|---|---|---|
| `require` | 33 | 16 |
| `autoload` | 1 | 22 |
| files loaded on normal boot | 35 | 23 |

The 16 remaining requires: 4 external (`doorkeeper`, `active_model`, `jwt`, `doorkeeper/request`) + 12 deliberate internal, each annotated with what would have to change to make it autoloadable.

## Verification

- `rake spec` + `rake eager_load` — green on all 7 gemfiles
- `rubocop` — no offenses
- Deferral confirmed empirically: 10 constants still pending after a normal boot


nbulaj (approved):
Same about changelog entry here please, otherwise LGTM ✅

55728:
Thank you for the review @nbulaj! 👀✨

PR #363, opened 2026-07-31
Omit `id_token` from an application-less password grant response

55728:
Follow-up to #349: the application-less `id_token` guard it added cannot reach the password grant, which is the one flow that actually issues an application-less token. That request still ends in an unhandled 500.

## The conditions

Three settings, all of them documented configuration, put a server on this path:

1. `password` is included in Doorkeeper's `grant_flows`
2. `Doorkeeper.config.skip_client_authentication_for_password_grant` is enabled
3. the token request carries `scope=openid`

Nothing here is a misconfiguration. Doorkeeper's own migration template calls out application-less access tokens as the expected consequence of the password grant:

```ruby
# Remove `null: false` if you are planning to use Password
# Credentials Grant flow that doesn't require an application.
t.references :application,    null: false
```

## Current behavior

With those three in place, `POST /oauth/token` with `grant_type=password` and no client credentials returns **HTTP 500**, and the access token row it created is **already committed** when it does.

The sequence:

1. Doorkeeper's `PasswordAccessTokenRequest#validate_client` passes a request that has neither a resolvable client nor credentials, because the skip option is on.
2. `before_successful_response` calls `find_or_create_access_token(client, ...)` with `client = nil`, so a token row with a `nil` `application_id` is committed.
3. This gem's `PasswordAccessTokenRequest#after_successful_response` presets `@response.id_token` whenever the `openid` scope is granted.
4. `TokenResponse#body` renders the preset ID Token, and `IdToken#as_json` raises `Errors::MissingRequiredClaim` for `aud` — which is sourced from the application's uid.

`Errors::MissingRequiredClaim` inherits from `OpenidConnectError`, not from `Doorkeeper::Errors::DoorkeeperError`, so Doorkeeper's `TokensController` does not rescue it (whether it uses the inline `rescue` of the 5.5 series or the later `rescue_from`). The exception reaches Rails, which answers 500 — not RFC 6749 §5.2's error response, and indistinguishable from a genuine server fault. Because the caller never receives the `access_token`, it cannot reuse the committed row either, so a client that retries leaves one orphan row per attempt (`reuse_access_token` is off by default).

## Why #349's guard does not reach this path

`TokenResponse#body` looks like it already covers this:

```ruby
id_token = self.id_token
id_token ||= build_id_token_for(token)
return super if id_token.nil?
```

`build_id_token_for` returns early when `token.application.blank?` — but only the second line calls it, and the second line only runs when nothing was preset. The password grant carries a nonce, so it presets an ID Token before the response is rendered; `self.id_token` is non-nil and `build_id_token_for` is never called. The guard covers the grants that arrive with nothing preset, which is how it reaches the refresh-token flow and misses this one.

The existing specs show the same seam from the other side. The application-less context in `spec/lib/oauth/token_response_spec.rb` has to cancel the preset before it can exercise the guard at all:

```ruby
context "with the openid scope present but no application" do
  let(:token) { create :access_token, application: nil, scopes: "openid email" }
  # See above: keep the outer `before` from eagerly building an IdToken.
  let(:id_token) { nil }
```

That `let` override is the guard's precondition written down: it holds for the flows that preset nothing, and the preset flows were never covered.

The authorization code flow presets an ID Token too, but `oauth_access_grants.application_id` is `null: false`, so it cannot produce an application-less token. The password grant is the only preset flow that reaches this state.

## The inconsistency this removes

On current master (after #349), the same application-less token is handled two different ways depending on which request touches it:

| Request | Application-less token + `openid` |
|---|---|
| password grant (issuing the token) | 500 `MissingRequiredClaim` |
| refresh-token flow (refreshing that same token) | 200, `id_token` omitted |

Refreshing the token succeeds while the request that created it fails. This PR makes the password grant answer the way the refresh already does, so it is not a new trade-off — it is the existing one, applied where it was missing.

## Why omitting `id_token` is conformant

OpenID Connect Core 1.0 defines the authorization code, implicit and hybrid flows. It does not define the resource owner password credentials grant, so that grant's token response is not an OIDC success response and is not required to carry an ID Token. None of the flows OIDC does define can produce an application-less token, so no conformant response loses its `id_token` here.

The response still echoes `"scope": "openid"` without an `id_token`, which is the same shape #349 settled on for the refresh-token flow.


nbulaj (approved):
LGTM ✅

55728:
Thanks for the review @nbulaj 😊

PR #368, opened 2026-08-08
Refresh the README

55728:
Three small accuracy fixes to the README, found while cross-checking it against current `master`.

>[!NOTE]
>Companion wiki updates (Migration guide for 2.0, issuer fallback, DCR parameter table) already landed on the wiki directly.


nbulaj (approved):
👍

PR #369, opened 2026-08-14
Add a Playwright end-to-end suite for the dummy app

55728:
## Motivation

The RSpec suite exercises controllers and requests in-process, but nothing ever drives the full OIDC flows through a real server over HTTP: there is currently no test that walks authorization → token exchange → userinfo end-to-end. Regressions that only surface at that level — middleware interactions, redirect handling, response modes, header-only error responses — stay invisible to CI even when every spec is green.

This PR adds a browser end-to-end suite that boots `spec/dummy` as a real Rails server and drives its dashboard UI with [Playwright](https://playwright.dev/). The dashboard already exercises discovery, authorize-URL building (nonce, PKCE S256 via WebCrypto, response modes), code exchange, ID token decoding, userinfo, introspection and revocation from plain JavaScript, so the suite automates exactly what a human would click through.

## What's covered (13 tests, chromium)

| Spec | Scenarios |
|------|-----------|
| `discovery` | openid-configuration contents; JWKS serves the RSA public key and never leaks private material |
| `auth-code` | code flow with nonce → token exchange → ID token claims (`iss`/`sub`/`aud`/`nonce`, response-targeted custom claims) → userinfo; reusing a consumed code fails with `invalid_grant` |
| `pkce` | S256 happy path; missing verifier → `invalid_request`, tampered verifier → `invalid_grant` |
| `implicit` | `id_token token` via fragment, `at_hash` present, fragment access token works against userinfo |
| `form-post` | `response_mode=form_post` delivers the code via POST and the exchange still succeeds |
| `token-management` | introspection `active: true` → revocation → `active: false`; revoked token gets a 401 from userinfo |
| `errors` | mismatched `redirect_uri` and unknown `client_id` render locally and never redirect to the foreign host |
| `consent` | the real Doorkeeper consent screen: approval issues an exchangeable code, denial returns `access_denied` |

## CI

A new independent `e2e` job runs the suite on two gemfiles — `rails_8.1` as the latest-stable signal and `doorkeeper_master` to catch upstream behavioral changes that only surface through real controllers and views. The existing build matrix and lint job are untouched. Node and the Chromium download are cached; traces, screenshots and the HTML report are uploaded as artifacts on failure.

The two gemfiles the job boots the dummy app with now include `puma`; the CI gemfiles never needed a server before because the RSpec suite runs in-process.

## Running locally

```sh
cd e2e
npm install
npx playwright install chromium
npx playwright test        # or: bundle exec rake e2e
```

`npx playwright test --ui` opens the interactive debugger. See the README's Development section and AGENTS.md for details.

<details>

<summary>Design decisions</summary>

- **Server lifecycle**: Playwright's `webServer` hook boots the dummy app in the development environment on port 3000 (the dashboard's default redirect URI) with `DATABASE_URL=sqlite3:db/e2e.sqlite3`, recreating the database on every run. The developer's `development.sqlite3` is never touched. Locally, an already running `rake server` is reused for fast iteration; CI always boots fresh.
- **No seeds**: each spec creates its own user and application through the dashboard forms, which doubles as coverage for the setup endpoints.
- **Consent screen**: two small dummy-app changes make it reachable. `skip_authorization` now yields when the new `force_consent` parameter is present, and `resource_owner_authenticator` remembers `params[:current_user]` in the session — Doorkeeper's consent form only replays the OAuth parameters, so without the session fallback the approval POST lost the resource owner and dead-ended at the nonexistent `/login`. Normal development URLs behave exactly as before, and the full RSpec suite is unaffected (443 examples, 0 failures).
- **Dashboard fix along the way**: body-less responses (e.g. the userinfo 401, whose error details live only in the `WWW-Authenticate` header) previously rendered a blank result; the dashboard now shows `HTTP <status>` plus that header.

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,68 @@ jobs:
       - name: Run tests
         timeout-minutes: 10
         run: bundle exec rake spec
+  e2e:
+    name: E2E (${{ matrix.gemfile }})
+    runs-on: ubuntu-latest
+    env:
+      CI: true
+      BUNDLE_GEMFILE: ${{ github.workspace }}/gemfiles/${{ matrix.gemfile }}.gemfile
+    if: |
+      !(   contains(github.event.pull_request.title,  '[ci skip]')
+        || contains(github.event.pull_request.title,  '[skip ci]'))
+    strategy:
+      fail-fast: false
+      matrix:
+        # Short names (not gemfile paths like the build matrix) so they can
+        # name the failure artifact below — expressions have no replace().
+        gemfile:
+          - rails_8.1
+          - doorkeeper_master
+    steps:
+      - name: Repo checkout
+        uses: actions/checkout@v7
+
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.4"
+          bundler-cache: true
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+          cache: npm
+          cache-dependency-path: e2e/package-lock.json
+
+      - name: Install Playwright
+        working-directory: e2e
+        run: npm ci
+
+      - name: Cache Playwright browsers
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ hashFiles('e2e/package-lock.json') }}
+
+      - name: Install Chromium
+        working-directory: e2e
+        run: npx playwright install chromium --with-deps
+
+      - name: Run E2E tests
+        timeout-minutes: 15
+        working-directory: e2e
+        run: npx playwright test
+
+      - name: Upload traces and report
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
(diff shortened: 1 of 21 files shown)


nbulaj:
Hello @55728 . Just curious why not Capybara with a modern engine ? :slightly_smiling_face:

55728:
Hello @nbulaj 😊

Honestly, the main reason is that I have more experience with Playwright from a personal project, so it was a natural first choice 😄

That said, I'm totally open to Capybara with a modern engine (e.g. Cuprite) if you'd prefer keeping things within the Ruby ecosystem. What do you think? 👀✨

nbulaj:
Very good question @55728 :smile: I don't have any experience with Playwright, but I know it's currently a modern approach. Capybara is far more easier for Ruby devs to support BTW. I'm trying to understand what will be the best option for the project and upcoming maintainers/contributors :thought_balloon:

55728:
Thanks @nbulaj! 😊 That makes total sense — keeping the barrier low for Ruby contributors is important. I’ll look into rewriting the suite with Capybara + Cuprite 👀✨

55728:
Replaced by #376

PR #370, opened 2026-08-14
Align the README with the wiki, and finish retiring `rake db:migrate`

55728:
## What was wrong

**The table of contents could not reach the Configuration section.** Every other entry is an in-page anchor; that one linked to the wiki, even though `## Configuration` existed a few lines below. The wiki link it carried was already the first thing that section offered.

**The post-logout migration steps were documented three times.** The `add_post_logout_redirect_uris` generator invocation appeared in the README, in [Migration from Old Versions](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/wiki/Migration-from-Old-Versions) — which the paragraph directly above it already links — and in [RP-Initiated Logout](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/wiki/RP-Initiated-Logout). The wiki copies carry the context the step needs; the README copy carried none.

**The list of wiki pages had fallen behind the wiki.** It named six pages; RP-Initiated Logout, Prompt and Max Age, Troubleshooting, Migration from Old Versions and both guides were missing. A hand-maintained index of a wiki that already has its own index only guarantees it drifts again with the next page.

**Two status entries had grown into implementation explanations** — which side provides the end-session endpoint, and which of the gem and Doorkeeper 6.0 serves the RFC 8414 document. Both are covered in more depth by the RP-Initiated Logout and Routes wiki pages.

**Nothing stated the supported versions.** The one question asked before adding a dependency could only be answered by opening the gemspec for Ruby and Doorkeeper, and the CI matrix for Rails, which the gemspec does not constrain at all.

**Every copy of the migration step named `rake db:migrate`.** Rails 5 moved `db:migrate` from `rake` to the `rails` command. The stale form appeared in the README twice, in the unreleased [#243] migration note, and — least visible but most likely to be copied straight into a terminal — in the error raised when `post_logout_redirect_uris` is written on an installation that has not run the upgrade migration:

```ruby
MISSING_COLUMN_MESSAGE =
  "can't write post_logout_redirect_uris: the oauth_applications column is missing — " \
  "run `rails generate doorkeeper:openid_connect:add_post_logout_redirect_uris` " \
  "and `rake db:migrate`"
```

The generator half immediately above it already says `rails`, so the two halves of one instruction disagreed.

## What this changes

- Table of contents points at `#configuration`, then at `#documentation` once the section is renamed
- The duplicated migration steps are dropped; the pointer above them now says upgrades may bring migrations with them
- The page list becomes a link to the wiki home, three entry points (Configuration, Migration from Old Versions, Troubleshooting) and a prose sentence covering the rest, so new wiki pages need no README change
- The status list goes back to naming specs and linking the wiki for the mechanics. The "client metadata" qualifier stays on the RP-Initiated Logout entry — it is what marks the support as partial
- A new Requirements table:

  | Dependency | Supported versions | Source |
  | --- | --- | --- |
  | Ruby | 3.2 or newer | gemspec |
  | Rails | 7.0 to 8.1 | CI matrix (the gemspec does not constrain Rails) |
  | Doorkeeper | 5.5 or newer, below 7.0 | gemspec |
  | ORM | Active Record only | `Doorkeeper::OpenidConnect.configure` |

  The section names the CI matrix as authoritative, since that is what actually decides whether a combination is supported.

- `rake db:migrate` → `rails db:migrate` in all four places the repository still named it: the two README occurrences, the [#243] migration note, and `MISSING_COLUMN_MESSAGE`. After this, `grep -rn 'rake db:migrate' README.md lib/` returns nothing
- Three further one-liners: the DiscoveryController and CHANGELOG links made absolute, because the gemspec ships the README and relative links point nowhere on rubygems.org; and a missing `sh` fence language


nbulaj (approved):
✅

PR #371, opened 2026-08-14
Refresh the CI references in CONTRIBUTING

55728:
`CONTRIBUTING.md` still tells contributors that their first line of defense is Travis CI:

> Our first line of defense is the [Travis CI](https://travis-ci.org/doorkeeper-gem/doorkeeper-openid_connect) build defined within [.travis.yml](.travis.yml) and triggered for every pull request.

There is no `.travis.yml` in the repository — that link is a 404 against our own tree — and the travis-ci.org build it names no longer exists. Meanwhile the three workflows a pull request actually faces are undocumented here, including the changelog verifier, which contributors currently discover only when it fails on them.

The paragraph now names them:

- [`ci.yml`](.github/workflows/ci.yml) — the specs, against each supported Ruby, Rails and Doorkeeper combination
- [`rubocop.yml`](.github/workflows/rubocop.yml) — style offenses, reported as review comments
- [`changelog.yml`](.github/workflows/changelog.yml) — expects `CHANGELOG.md` to be updated unless the pull request is labeled `Skip-Changelog`

A second commit changes the branch snippet from `git checkout -B feature/contributing` to `-b`. `-B` resets an existing branch of that name to the current commit, so a contributor who runs the documented command a second time silently discards whatever was already on that branch; `-b` refuses instead, which is what the instruction means.


nbulaj (approved):
Nice 👍

nbulaj:
BTW I think we have to adjust changelog rules to ignore changes in *.md files. I'm not sure how valuable such changelogs for end users which later checks them for any useful info about upgrades. It looks more like an internal stuff. I'd prefer to concentrate on real library API changes when checking them rather than reading Updated readme / Fixed CI / whatever else doesn't affect end users directly. WDYT (as a **separate thing for sure**, it's just a side note)?

55728:
+1 on adjusting the changelog rules 👍

One thing to keep in mind: external contributors can't add labels themselves, so a label-based skip (like `Skip-Changelog`) still requires maintainer intervention before CI goes green. That makes the contributor experience a bit bumpy.

A workflow-level approach — e.g. checking the changed file paths and skipping when only `*.md` files are touched — would let the CI pass on its own without any extra steps from either side. Happy to open a follow-up PR for that if you'd like!

55728:
I also should rewrite my entries on CHANGELOG.md, which are a bit long, before the next release 😓

nbulaj:
I'm for a workflow-level approach, it will be simpler for contributors and they don't need to manage any labels and so on (also simpler for maintainers as well).

55728:
Agree 👍 I will open a follow-up PR 🚀

PR #372, opened 2026-08-14
Fix String/Symbol duplicates in `claims_supported`

55728:
## Problem

`claims_supported` in the discovery document is built as

```ruby
%w[iss sub aud exp iat] | openid_connect.claims.to_h.keys
```

The five base claims are Strings, but the configured claim names are Symbols (the claims builder stores them in an `OpenStruct`, which symbolizes its keys), and `Array#|` never treats `"sub"` and `:sub` as the same element. Nothing stops a host application from configuring a claim named after a base claim, so with

```ruby
claims do
  claim(:sub) { |resource_owner| resource_owner.some_custom_subject }
end
```

the discovery document advertises `sub` twice:

```json
"claims_supported": ["iss", "sub", "aud", "exp", "iat", "sub", ...]
```

Since the helper moved into `DiscoveryHelpersMixin`, the same duplicate also reaches the RFC 8414 metadata document that Doorkeeper 6.0 serves through `MetadataExtension` — both documents are affected, from one line.

## Fix

Stringify the configured claim names before the union so the deduplication works:

```ruby
%w[iss sub aud exp iat] | openid_connect.claims.to_h.keys.map(&:to_s)
```

For claims that don't collide with a base claim the rendered JSON is unchanged — the Symbols already serialized as strings; only the Ruby-side array becomes consistently typed.


lib/doorkeeper/openid_connect/discovery_helpers_mixin.rb
```diff
@@ -16,7 +16,9 @@ def issuer
       end
 
       def claims_supported(openid_connect)
-        %w[iss sub aud exp iat] | openid_connect.claims.to_h.keys
+        # The configured claim names are Symbols; align the types so a custom
+        # claim shadowing a base claim is deduplicated by the union.
+        %w[iss sub aud exp iat] | openid_connect.claims.to_h.keys.map(&:to_s)
```
nbulaj: Can we extract it into a BASE_CLAIMS and freeze to avoid memory/object allocations on each calls please?

nbulaj (approved):
Just minor advise, otherwise LGTM 👍

lib/doorkeeper/openid_connect/discovery_helpers_mixin.rb
55728: Thanks 😊 Extracted it as a frozen `BASE_CLAIMS` constant, so the array and its five strings are allocated once at load time instead of on every discovery request. The union still returns a new array, so nothing can write through to the constant.

PR #373, opened 2026-08-14
Answer `prompt=consent` with JSON when Doorkeeper runs in API mode

55728:
Under Doorkeeper's `api_only` configuration, an authorization request carrying `prompt=consent` answers **`200 text/plain` with an empty body**, and raises nothing. The relying party is told its request succeeded and receives nothing it can act on.

## The mechanism

`handle_oidc_prompt_consent!` renders Doorkeeper's `:new` template directly:

```ruby
def handle_oidc_prompt_consent!(owner)
  return unless owner

  clear_oidc_response
  render :new
end
```

With `api_only`, `Doorkeeper::ApplicationController` descends from `ActionController::API` (`Doorkeeper.config.resolve_controller(:base)`), which carries no template rendering at all. `render :new` there does not raise `ActionView::MissingTemplate` — there is no view layer to raise it — it simply produces an empty 200. `prompt=consent` is a standard [OIDC Core §3.1.2.1](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) parameter, so any relying party can reach this.

## The fix

Present the pre-authorization as JSON in that mode, which is what Doorkeeper's own `AuthorizationsController#render_success` does for the consent step under `api_only`, so the two agree on what an API-mode consent step looks like:

```ruby
if Doorkeeper.configuration.api_only
  render json: pre_auth
else
  render :new
end
```

After the change, `prompt=consent` answers `200 application/json` with the same pre-authorization document the no-`prompt` case returns.


nbulaj (approved):
LGTM ✅

55728:
Thanks for the review 😊

PR #367, opened 2026-08-08
CI: exercise the released Rails 8.1 series

55728:
The build matrix stops at `gemfiles/rails_8.0.gemfile`, but the Rails 8.1 series has been out for a while (latest: 8.1.3.1). This adds it to CI so 2.0 ships with the current Rails release actually exercised.

PR #374, opened 2026-08-14
Tighten the Unreleased changelog entries

55728:
Follow-up to the side note in https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/371#issuecomment-5291466469

The Unreleased section had grown to 33 entries, many of which read more like implementation notes than upgrade guidance. This trims them with two rules:

1. **Drop entries invisible to upgraders** — CI pins, spec additions, doc-only edits, behavior-preserving refactors, and fixes for unreleased regressions (#342, #346, #354, #356, #357, #366, #367, #371). #370 is partially kept because it corrects the `rake db:migrate` command in an error message.

2. **Shorten the rest to state the _what_ without the _how_** — RFC section numbers, internal class names, and implementation paths are trimmed; the PR number still links to the full context.

Breaking and security entries keep their detail since they ask readers to act.

No code change.


nbulaj (approved):
That really makes sense 👍 

BTW have we moved (or already have) these important notes somewhere (upgrade guides, whatever visible for developers) ?

55728:
Thanks for the review 😊

Most of the breaking changes already have entries on the [Migration from Old Versions](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/wiki/Migration-from-Old-Versions) wiki page, but I'll do a pass to make sure nothing is missing before the release 👍

PR #375, opened 2026-08-14
Skip the changelog check for documentation-only pull requests

55728:
Follow-up to https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/371#issuecomment-5291466469

The changelog verifier asks every pull request to update `CHANGELOG.md`. The escape hatch today is the `Skip-Changelog` label, which an external contributor cannot apply to their own pull request — so a documentation-only PR stays red until a maintainer intervenes.

This adds a step that checks the changed file paths before the enforcer runs. When every changed file ends in `.md`, the enforcer is skipped automatically. The `Skip-Changelog` and `dependencies` labels still work as manual escape hatches for everything else.

The check uses step-level `if` rather than `paths-ignore` on the workflow trigger — `paths-ignore` would prevent the job from firing at all, which blocks merge when the check is required.

No library change.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -16,7 +16,33 @@ jobs:
   changelog:
     runs-on: ubuntu-latest
     steps:
+      # A pull request that only edits Markdown files documents the project
+      # instead of changing the library, so it leaves an upgrader nothing to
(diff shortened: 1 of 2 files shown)


nbulaj (approved):
Looks good 👍

55728:
Thank you so much 😊

PR #376, opened 2026-08-14
Add a Capybara end-to-end suite for the dummy app

55728:
Supersedes #369 — rewritten from Playwright to Capybara + Cuprite per [the discussion there](https://github.com/doorkeeper-gem/doorkeeper-openid_connect/pull/369#issuecomment-5292276439), keeping the test stack within the Ruby ecosystem so any Ruby contributor can pick up and maintain it.

### Motivation

The RSpec suite exercises controllers and requests in-process, but nothing ever drives the full OIDC flows through a real server over HTTP: there is currently no test that walks authorization → token exchange → userinfo end-to-end. Regressions that only surface at that level — middleware interactions, redirect handling, response modes, header-only error responses — stay invisible to CI even when every spec is green.

This PR adds a browser end-to-end suite that boots `spec/dummy` as a real Rails server and drives its dashboard UI with [Capybara](https://github.com/teamcapybara/capybara) + [Cuprite](https://github.com/rubycdp/cuprite) (Ferrum/CDP, no Selenium). No Node.js required — just an installed Chrome.

### What's covered (13 tests)

| Spec             | Scenarios                                                    |
| ---------------- | ------------------------------------------------------------ |
| discovery        | openid-configuration contents; JWKS serves the RSA public key and never leaks private material |
| auth-code        | code flow with nonce → token exchange → ID token claims (`iss`/`sub`/`aud`/`nonce`, response-targeted custom claims) → userinfo; reusing a consumed code fails with `invalid_grant` |
| pkce             | S256 happy path; missing verifier → `invalid_request`, tampered verifier → `invalid_grant` |
| implicit         | `id_token token` via fragment, `at_hash` present, fragment access token works against userinfo |
| form-post        | `response_mode=form_post` delivers the code via POST and the exchange still succeeds |
| token-management | introspection `active: true` → revocation → `active: false`; revoked token gets a 401 from userinfo |
| errors           | mismatched `redirect_uri` and unknown `client_id` render locally and never redirect to the foreign host |
| consent          | the real Doorkeeper consent screen: approval issues an exchangeable code, denial returns `access_denied` |

### Running locally

```sh
bundle exec rake e2e
```

`HEADLESS=false` opens a visible browser. The suite boots the app in the development environment against its own `spec/dummy/db/e2e.sqlite3`, so it leaves the one behind `bundle exec rake server` alone.

### Design decisions

<details> <summary>Why development environment, not test?</summary>

The dashboard needs two things that only work in development: Doorkeeper's `skip_authorization` (which the consent specs override with `force_consent=1`) and relaxed `force_ssl_in_redirect_uri` for `http://` callback URIs. Switching the existing test environment would risk breaking the 444 in-process specs, so the E2E suite runs in a separate process with its own database.

</details> <details> <summary>Why a separate rake task?</summary>

`rake spec` excludes `spec/e2e/` and `rake e2e` targets only `spec/e2e/`. One process can only boot the dummy app in one environment, so they cannot share a run. Running bare `rspec` without the task prints an explicit message pointing to `rake e2e`.

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,46 @@ jobs:
       - name: Run tests
         timeout-minutes: 10
         run: bundle exec rake spec
+  e2e:
+    name: E2E (${{ matrix.gemfile }})
+    runs-on: ubuntu-latest
+    env:
+      CI: true
+      BUNDLE_GEMFILE: ${{ github.workspace }}/gemfiles/${{ matrix.gemfile }}.gemfile
+    if: |
+      !(   contains(github.event.pull_request.title,  '[ci skip]')
+        || contains(github.event.pull_request.title,  '[skip ci]'))
+    strategy:
+      fail-fast: false
+      matrix:
+        # Short names (not gemfile paths like the build matrix) so they can
+        # name the failure artifact below — expressions have no replace().
+        gemfile:
+          - rails_8.1
+          - doorkeeper_master
+    steps:
+      - name: Repo checkout
+        uses: actions/checkout@v7
+
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "3.4"
+          bundler-cache: true
+
+      # Cuprite drives the Chrome the runner image already ships, so the suite
+      # needs nothing beyond the bundle.
+      - name: Run E2E tests
+        timeout-minutes: 15
+        run: bundle exec rake e2e
+
+      - name: Upload the pages of failed examples
+        if: failure()
+        uses: actions/upload-artifact@v4
+        with:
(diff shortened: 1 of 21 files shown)


nbulaj (approved):
Damn this is rock solid 🚀 🤘

nbulaj:
I'm totally :+1: with a separate rake task and the current implementation, sharing the same vision! Thanks

55728:
Thank you for the review and the kind words @nbulaj! 🚀

The journey from the Playwright approach in #369 to this Capybara suite was a great learning experience — your feedback helped shape a solution that fits naturally into the existing test infrastructure.

Glad we're on the same page with the separate rake task 😊🤘

PR #381, opened 2026-08-15
Echo the registered `client_name` in the Dynamic Client Registration response

55728:
[RFC 7591 §3.2.1](https://www.rfc-editor.org/rfc/rfc7591#section-3.2.1) requires a successful registration response to include **all registered client metadata**:

> Additionally, the authorization server MUST return all registered metadata about this client, including any fields provisioned by the authorization server itself.

The registration endpoint accepts `client_name` and stores it as the application's `name`, but the `201 Created` response never echoes it back — the client has no confirmation of what was registered. Since the response is also the RFC's mechanism for telling the client which values the server actually recorded (clients are expected to replace their own values with the returned ones), silently dropping an accepted field breaks that contract.

>[!NOTE]
>
> #### Why the echo is unconditional
> 
> `client_name` is always present on a successfully created application: the application model validates `name` for presence, so a registration request without `client_name` already fails with `invalid_client_metadata` before reaching the response builder. There is no "registered without a name" state to guard against, unlike `post_logout_redirect_uris` (which stays conditional because registration of that field is optional).


nbulaj (approved):
✅

PR #391, opened 2026-08-23
Fix the remaining RSpec/ContextWording, StubbedMock and DescribeMethod offenses

55728:

Nine offenses, all in specs:

- **ContextWording (6):** reword context descriptions to start with when/with/without (`"and a matching token"` → `"with a matching token"`, `"token reuse"` → `"with reuse_access_token enabled"`, …)
- **StubbedMock (2):** `expect(::Rails.env).to receive(:production?)` → `allow`; the examples use the stubbed return value and never asserted the call
- **DescribeMethod (1):** `describe Doorkeeper::OpenidConnect, "configuration"` → `".configuration"`, the method the subject calls

Then the three cops are dropped from `.rubocop_todo.yml`.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -26,14 +26,3 @@ Style/SymbolProc:
 
 RSpec/SubjectStub:
   Enabled: false
-
-# Configuration parameters: Prefixes, AllowedPatterns.
-# Prefixes: when, with, without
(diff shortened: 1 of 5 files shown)


nbulaj (approved):
Love it ✅

55728:

PR #392, opened 2026-08-23
Keep RSpec/SubjectStub enabled and mark its legitimate uses inline

55728:

The todo file disabled `RSpec/SubjectStub` globally. The 17 offenses it hid all stub a **collaborator the method under test reads from**, never the method being asserted:

| spec | stubbed | why |
|---|---|---|
| `authorizations_controller_spec` #reauthenticate / #select_account | `clear_oidc_response`, `performed?` | isolate the method's own branching from the helpers it delegates to |
| `authorizations_controller_spec` #clear_oidc_response | `response_body=` | observe the reset without rendering |
| `id_token_spec` / `user_info_spec` #as_json | `claims` | feed blank claims directly to the serializer |
| `application_spec` | `has_attribute?` | emulate an installation that has not run the migration |
| `token_endpoint_auth_methods_supported_mixin_spec` | `doorkeeper_config` | emulate each Doorkeeper generation (the mixin's only input) |

Those are partial doubles used to isolate a method, which is not the smell the cop targets. So instead of disabling the cop, wrap the seven example groups in `rubocop:disable/enable` with a one-line reason each and drop the entry from the todo file, leaving the cop enabled for new code.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -23,6 +23,3 @@ Style/SymbolProc:
     - "spec/lib/id_token_spec.rb"
     - "spec/lib/user_info_spec.rb"
     - "spec/support/doorkeeper_configuration.rb"
-
-RSpec/SubjectStub:
-  Enabled: false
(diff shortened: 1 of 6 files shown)


nbulaj (approved):
I'm OK with it ✅

PR #393, opened 2026-08-23
Exclude spec/ from Metrics/MethodLength and Metrics/AbcSize

55728:

`Metrics/BlockLength` and `Metrics/ParameterLists` already skip `spec/`; spec support code and the dummy app favour one readable setup method over a split one for the same reason. This widens the existing `spec/dummy/db/**/*` exclusion to the whole tree, so every remaining Metrics offense behind the todo file is in library code.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -53,9 +53,14 @@ Metrics/BlockLength:
 Metrics/ParameterLists:
   Exclude:
     - spec/**/*
+# Spec support code and the dummy app favour one readable setup method over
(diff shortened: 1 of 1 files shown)


nbulaj (approved):
Does `Metrics/BlockLength:` already there as well? I mean ignored for `spec/` directory (since we're using Rspec). If not I believe we have to add it

55728:
Yes, `Metrics/BlockLength` already excludes `spec/**/*` (since 17ab5e6). `Metrics/ParameterLists` too.

PR #395, opened 2026-08-23
Bring nine library methods under Metrics/MethodLength and AbcSize

55728:

Each commit touches one method and can be reviewed on its own:

| commit | method | change |
|---|---|---|
| 1 | `OpenidConnect.call_issuer` | fold the 4-way arity dispatch onto `when n then …` lines |
| 2 | `Prompt#apply_oidc_prompt!` | same for the prompt-value dispatch; the `create` note stays |
| 3 | `Prompt#return_without_oidc_prompt_param` | extract `query_parameters_without_oidc_prompt` (query rewrite vs URL assembly) |
| 4 | `TokenMatching#oidc_subset_token_exists?` | `return true` from the batch block replaces the `match_found` flag + `break` |
| 5 | `Config::Builder#resolve_validated_model` | extract `validate_model_methods!` |
| 6 | `Rails::Routes#generate_routes!` | extract `map_oauth_routes` and `well_known_scope` (the namespaced-mount note moves with it) |
| 7 | `DynamicClientRegistrationController#register` | extract `record_invalid_response` for the rescue body |
| 8 | `ClaimsBuilder.generate` | extract `claim_applies?` (scope-and-response predicate) |
| 9 | `ErrorResponse#handle_oidc_error!` | both branches passed the same five kwargs and differed only in class → pick the class in `oidc_error_response_class`, build once |

No spec was changed. Every commit passes RuboCop individually. `rake spec`: 455/0 (default), 453/0 (doorkeeper 5.5), 464/0 (doorkeeper master), 455/0 (rails 7.0).

### DIFF
--- a/app/controllers/doorkeeper/openid_connect/dynamic_client_registration_controller.rb
+++ b/app/controllers/doorkeeper/openid_connect/dynamic_client_registration_controller.rb
@@ -16,10 +16,7 @@ module Doorkeeper
         client = Doorkeeper.configuration.application_model.create!(application_params(registration))
         render json: registration_response(client, registration), status: :created
       rescue ActiveRecord::RecordInvalid => e
-        render json: {
-          error: registration_error_code(e.record),
-          error_description: e.record.errors.full_messages.join(", "),
-        }, status: :bad_request
+        render json: record_invalid_response(e.record), status: :bad_request
       end
 
       private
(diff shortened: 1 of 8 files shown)


app/controllers/doorkeeper/openid_connect/dynamic_client_registration_controller.rb
```diff
@@ -51,6 +48,13 @@ def authorized?(authorizer)
 
       # RFC 7591 §3.2.2 registration error codes: `invalid_redirect_uri` for
```
nbulaj: I believe this comment has to be before `registration_error_code` 👀

55728: Thanks for the sharp eye! 👀✨ Moved the extracted method above the comment so the comment sits right on `registration_error_code` again.

nbulaj (approved):
✅

PR #396, opened 2026-08-23
Drop the Metrics overrides from .rubocop_todo.yml

55728:

After the previous PRs only two `Metrics/AbcSize` offenses remain, both in methods that build a metadata document: `DiscoveryController#provider_response` and `DynamicClientRegistrationController#registration_response` (RFC 7591 §3.2.1). Splitting them would hide the shape of the published document, so they get an inline `rubocop:disable Metrics/AbcSize` with the reason, and `Metrics/MethodLength` / `Metrics/AbcSize` now run at RuboCop's defaults everywhere else.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -6,13 +6,6 @@
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
-Metrics/MethodLength:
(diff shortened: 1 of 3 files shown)


nbulaj (approved):
LGTM ✅

55728:
Thank you! 👀✨

